Issue #73, opened 2019-07-26
add kicktimes to kick_plane.tfs


mihofer:
add a column with kick times (LHC sdds timeformat) to kick files 
potentially also put both planes in one file again

JoschD:
Why have these been removed in the first place? @lmalina

JoschD:
Fixed by #77

PR #77, opened 2019-07-31
#73 Add Kicktime to kick_$.tfs

JoschD:
* Refactored kick.py to use only constants as Column names
  * harpy/handler.py: added TIME to headers.
  * accelerator.py: refactored strings of element types to AccElementTypes class

### DIFF
--- a/omc3/definitions/formats.py
+++ b/omc3/definitions/formats.py
@@ -1,0 +1,1 @@
+TIME = "%y_%m_%d@%H_%M_%S"  # CERN default
--- a/omc3/harpy/handler.py
+++ b/omc3/harpy/handler.py
@@ -12,6 +12,7 @@ import numpy as np
 import pandas as pd
 
 import tfs
+from definitions import formats
 from utils.contexts import timeit
 from utils import logging_tools
 from harpy import frequency, clean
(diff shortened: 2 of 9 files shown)


mihofer:
https://github.com/pylhc/omc3/blob/4ad993aaec370c9fab2e1b14d396a31c27d5a8a0/omc3/harpy/handler.py#L145-L158

why not use instead of just date as argument a dict containing additional headers, would allow to easily implement more headers at a later point if needed

mihofer:
https://github.com/pylhc/omc3/blob/4ad993aaec370c9fab2e1b14d396a31c27d5a8a0/omc3/optics_measurements/kick.py#L47-L50

maybe change to RESCALE_FACTOR{plane}, kick files could then be merged without name clash

mihofer:
should optic_measurements/constants.py be put in definitions as other scripts (kmod, amp. det script, ..) should also rely on the conventions established there?

JoschD:
> https://github.com/pylhc/omc3/blob/4ad993aaec370c9fab2e1b14d396a31c27d5a8a0/omc3/harpy/handler.py#L145-L158
> 
> why not use instead of just date as argument a dict containing additional headers, would allow to easily implement more headers at a later point if needed

A dict would make sense if the headers are added automatically. But as there is some formatting going on, one has to handle each entry individually anyway (most likely). Changing `date` to `addition_headers` later if needed should be no problem though.

JoschD:
> should optic_measurements/constants.py be put in definitions as other scripts (kmod, amp. det script, ..) should also rely on the conventions established there?

Yes... but we need come to some convetion later. New issue, I guess

JoschD:
> https://github.com/pylhc/omc3/blob/4ad993aaec370c9fab2e1b14d396a31c27d5a8a0/omc3/optics_measurements/kick.py#L47-L50
> 
> maybe change to RESCALE_FACTOR{plane}, kick files could then be merged without name clash

Maybe. Also new issue to talk about.

PR #90, opened 2019-08-06
Enhancement/81/ptc

awegsche:
Added a PTC_trackone reader.
Enables us to check with PTC simulations which is needed for further development of algorithms.


JoschD:
Further suggestions appreciated

lmalina:
closing: #91 is more advanced.

PR #91, opened 2019-08-07
Enhancement/81/ptc josch

JoschD:
Modified version from Andreas suggestion


JoschD:
Further suggestions appreciated

PR #94, opened 2019-08-13
Enhancement/93/rdt freq header

mihofer:
RDT functionality and subsequently also this enhancement is currently not tested, should be thought about

also minor correction in measure_optics is included


mihofer:
line and frequency determination is now unified in function (could be nicer though),
line could also be added to header, however a previous conversion to string or addition of list writing in tfs would be needed

lmalina:
The test coverage will be improved with PTC based tests

PR #113, opened 2019-10-14
added different beam directions for LER and HER ring in SKEKB class

JacquelineKeintzel:
Assigned different beam directions for HER (-1) and LER (+1) ring in SKEKB, similar as for Beam 1 and Beam 2 in LHC.


omc3/model/accelerators/skekb.py
```diff
@@ -21,5 +24,37 @@ def verify_object(self):  # TODO: Maybe more checks?
             raise AcceleratorDefinitionError("SuperKEKB doesn't have a model creation, calling it this "
                                              "way is most probably wrong.")
 
-    def get_beam_direction(self):
+    @staticmethod
+    def get_class_parameters():
+        params = EntryPointParameters()
+        params.add_parameter(flags=["--ring"], help="HER or LER ring.", name="ring", type=str)
```
lmalina: I would add choices=("ler", "her") , this will also clarify the condition in line 44

PR #114, opened 2019-10-14
Adding model creators and sequence files

lmalina:
Since it will take time to merge ABP and OP sequence repositories and we need this, it should be merged. There are tests missing for the moment as the Beta-Beat GUI is needed to create the models.


lmalina:
I guess you mean the generated models, right? They are the same. These codes are actually what we use in hack_for_gui branch. In other words, we already use these codes in the GUI. 
The codes could be nicer though, but that is for later when we'll have everything in Python and write tests.

PR #117, opened 2019-10-16
Enhancement/38/separate packages

lmalina:
Replaced tfs, sdds, and generic_parser by separated packages
    packages are added to requirements file
    entrypoints are adapted, i.e. keyword "flags" is not needed
    cleaned up a bit the options of hole_in_one
        removed max_beta_beating and max_closed_orbit


omc3/madx_wrapper.py
```diff
@@ -13,7 +13,7 @@
 import contextlib
 import warnings
 from tempfile import mkstemp
-from generic_parser.entrypoint import entrypoint, EntryPointParameters
+from generic_parser.entrypoint_parser import entrypoint, EntryPointParameters
```
JoschD: this can be imported directly from generic_parser, btw

omc3/model/accelerators/lhc.py
```diff
@@ -8,7 +8,7 @@
 from model.accelerators.accelerator import Accelerator, AcceleratorDefinitionError, AccExcitationMode, AccElementTypes
 from utils import logging_tools
 import tfs
-from generic_parser.entrypoint import EntryPointParameters
+from generic_parser.entrypoint_parser import EntryPointParameters
```
JoschD: same here

PR #120, opened 2019-10-21
Enhancement/118/multiple noise level

### LINKED ISSUE #118: add option in tbt_converter to run with multiple noise level
extends #100 
currently only one noise level can be specified 
with list as input different noise level files can be generated in one go

mihofer:
fixes #118 , new produced filescheme is basename_nNOISE_rREP.sdds
no specific test for added noise option (as before), 
to check probably need to modify _compare_tbt function to take as input either seed for rnd generator or just precision option

### DIFF
--- a/omc3/tbt/handler.py
+++ b/omc3/tbt/handler.py
@@ -74,7 +74,7 @@ def _matrices_to_array(tbt_data):
 def _add_noise(data, noise):
     if noise <= 0.0:
         return data
-    return data + noise * np.random.randn(data.shape)
+    return data + noise * np.random.standard_normal(data.shape)
 
 
 def write_lhc_ascii(output_path, tbt_data):
--- a/omc3/tbt/handler.py
+++ b/omc3/tbt/handler.py
@@ -72,8 +72,6 @@ def _matrices_to_array(tbt_data):
 
 
 def _add_noise(data, noise):
-    if noise <= 0.0:
-        return data
     return data + noise * np.random.standard_normal(data.shape)
 
 
--- a/omc3/tbt_converter.py
+++ b/omc3/tbt_converter.py
@@ -34,7 +34,7 @@ def converter_params():
                          help="Choose the datatype from which to import. ")
     params.add_parameter(name="realizations", type=int, default=1,
                          help="Number of copies with added noise")
-    params.add_parameter(name="noise_levels", nargs='+', type=float,
+    params.add_parameter(name="noise_levels", nargs='+', type=float, default=[None],
                          help="Sigma of added Gaussian noise")
     return params
 
@@ -70,7 +70,6 @@ def converter_entrypoint(opt):
     """
     if opt.realizations < 1:
         raise ValueError("Number of realizations lower than 1.")
-    opt.noise_levels = [None] if opt.noise_levels is None else opt.noise_levels
     iotools.create_dirs(opt.outputdir)
     save_options_to_config(join(opt.outputdir, DEFAULT_CONFIG_FILENAME.format(
         time=datetime.utcnow().strftime(formats.TIME))), OrderedDict(sorted(opt.items())))
--- a/omc3/tbt_converter.py
+++ b/omc3/tbt_converter.py
@@ -34,7 +34,7 @@ def converter_params():
                          help="Choose the datatype from which to import. ")
     params.add_parameter(name="replications", type=int, default=1,
                          help="Number of copies with added noise")
-    params.add_parameter(name="noise_to_add", type=float, default=0.0,
+    params.add_parameter(name="noise_to_add", nargs='*', default=[0.0],
                          help="Sigma of added Gaussian noise")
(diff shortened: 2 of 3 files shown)


omc3/tbt_converter.py
```diff
@@ -79,10 +79,13 @@ def converter_entrypoint(opt):
 def _read_and_write_files(opt):
     for input_file in opt.files:
         tbt_data = tbt.read_tbt(input_file, datatype=opt.tbt_datatype)
-        for i in range(opt.replications):
-            suffix = f"_{i}" if opt.replications > 1 else ""
-            tbt.write(join(opt.outputdir, f"{_file_name(input_file)}{suffix}"), tbt_data,
-                      noise=opt.noise_to_add)
+        for n in opt.noise_to_add:
```
JoschD: As you allow empty lists with `nargs='*'` I would distinguish here between noise given and not given. The way you have it at the moment will, for example, not add any noise-level into the filename, if you give `noise_to_add = [10.]` even though there is clearly noise

omc3/tbt_converter.py
```diff
@@ -34,7 +34,7 @@ def converter_params():
                          help="Choose the datatype from which to import. ")
     params.add_parameter(name="replications", type=int, default=1,
                          help="Number of copies with added noise")
-    params.add_parameter(name="noise_to_add", type=float, default=0.0,
+    params.add_parameter(name="noise_to_add", nargs='*', default=[0.0],
```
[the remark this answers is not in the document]
JoschD: Also, personal opinion, the variables should be named differently (also to make it immediately clear they are both related to the noise):
`noise_to_add` -> `noise_levels`
`replications` ->  `noise_realizations`

omc3/tbt/handler.py
```diff
@@ -74,7 +74,7 @@ def _matrices_to_array(tbt_data):
 def _add_noise(data, noise):
     if noise <= 0.0:
```
JoschD: looking at the rest of the code, I think this should read `if noise is None`

omc3/tbt_converter.py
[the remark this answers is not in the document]
JoschD: Related: You should not have replications if you don't have any noise set.

omc3/tbt/handler.py
lmalina: the check for None is outside of the method, we can remove the condition completely, thus allowing for negative numbers as well as @JoschD mentioned.

omc3/tbt_converter.py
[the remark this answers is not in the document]
lmalina: I would keep type=float and set nargs='+', with no default, later we check whether it is None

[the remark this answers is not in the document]
lmalina: What about the changes to noise_levels and keep replications?

omc3/tbt_converter.py
lmalina: That was initially my suggestion, but you are right: if there is noise level specified, i.e. not None, should be in the name.

lmalina: This part can then be a loop over replications, with a check if noise is None inside it...either writes one file or comprehends the writing over list of noise levels...

omc3/tbt_converter.py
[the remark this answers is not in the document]
JoschD: > What about the changes to noise_levels and keep replications?

`replications` is a horrible name ^^

omc3/tbt_converter.py
```diff
@@ -58,18 +58,20 @@ def converter_entrypoint(opt):
...
-        Flags: **--noise_to_add**
-        Default: ``0.0``
+        Flags: **--noise_levels**
+        Default: None
 
     """
-    if opt.replications < 1:
-        raise ValueError("Number of replications lower than 1.")
+    if opt.noise_levels is None: 
+        raise ValueError("No noise level specified.")
```
lmalina: We should allow for no noise specified.

omc3/tbt_converter.py
```diff
@@ -79,10 +81,13 @@ def converter_entrypoint(opt):
...
-            suffix = f"_{i}" if opt.replications > 1 else ""
-            tbt.write(join(opt.outputdir, f"{_file_name(input_file)}{suffix}"), tbt_data,
-                      noise=opt.noise_to_add)
+        for i in range(opt.realizations):
+           suffix = f"_r{i}" if opt.realizations > 1 else ""
+           for n in opt.noise_levels:
+             noise_suffix = f"_n{n}" if not n == 0 else ""
+             tbt.write(join(opt.outputdir, f"{_file_name(input_file)}{noise_suffix}{suffix}"),
+                         tbt_data=tbt_data,
+                         noise=n)
```
lmalina: Influenced by possible None

lmalina: The indentation doesn't seem right

Issue #124, opened 2019-11-12
implement average tbtdata option


mihofer:
in multiparticle trackone data, oftimes only the c.o.g. is of interest, currently requires preprocessing
alternative: implement average tbt function which takes multiparticle TbtData and returns single particle Tbtdata with average position

lmalina:
Fixed by #125

PR #125, opened 2019-11-12
Minorissue/123/trackone all obs points

### LINKED ISSUE #123: trackone reader only allows for BPM
reader_trackone only takes observations points which contain 'BPM', which appears to be an overly stringent condition

mihofer:
Closes #123 & #124


mihofer:
and fixes #126

omc3/tbt/handler.py
```diff
@@ -32,6 +32,28 @@ def __init__(self, matrices, date, bunch_ids, nturns):
...
+
+    matrices = [{plane: pd.DataFrame(index=bpm_names,
+                                     data=get_averaged_data(bpm_names, data, plane, tbtdata.nturns),
+                                     dtype=float) for plane in PLANES}]
+    return TbtData(matrices, tbtdata.date, [1], tbtdata.nturns)
+
+
+def get_averaged_data(bpm_names, data, plane, turns):
+
+    bpm_data = np.zeros((len(bpm_names), len(data), turns))
```
lmalina: This will not work if some of the particles are lost

omc3/tbt/reader_trackone.py
```diff
@@ -69,12 +71,8 @@ def get_structure_from_trackone(nturns=0, npart=0, infile='trackone'):
             parts = l.split()
             if parts[0] == '#segment':
                 bpm_name = parts[-1].upper()
-                if ('BPM' in bpm_name) and (bpm_name not in bpms.keys()):
+                if (np.all([k not in bpm_name.lower() for k in ['start', 'end']])) and (bpm_name not in bpms.keys()):
                     bpms[bpm_name] = np.empty([npart, nturns, 8], dtype=float)
-            elif 'BPM' in bpm_name:
-                bpms[bpm_name][int(parts[0]) - 1, int(parts[1]) - 1, :] = np.array(parts[2:])
+            elif (np.all([k not in bpm_name.lower() for k in ['start', 'end']])):
+                bpms[bpm_name][np.abs(int(float(parts[0]))) - 1, int(float(parts[1])) - 1, :] = np.array(parts[2:])
```
lmalina: why the absolute value?

mihofer: in case you use the only_average option in madx, the particle id is saved as -1, in order to make indexing work the abs is introduced

omc3/tbt/handler.py
mihofer: fixed with last commit

Issue #134, opened 2019-12-04
Add response matrix calcuation


lmalina:
Needed for global corrections
also it is not possible to use ```pandas.Panel```, which not present anymore in required version of pandas

PR #144, opened 2019-12-19
Madx macros model creators and accelerator classes from #134

lmalina:
Refactors accelerator class and removes most of compound class structures 
Reduces madx macros used in to create models and removes madx templates
Adds more compact LHC b2 errors
There are no more hard-coded links to outside of the repository
LHC model creation is refactored accordingly


omc3/model/accelerators/psbooster.py
```diff
@@ -4,63 +4,59 @@
...
 
     @classmethod
     def get_file(cls, filename):
-        return os.path.join(CURRENT_DIR, "psbooster", filename)
+        return os.path.join(CURRENT_DIR, cls.NAME, filename)
 
-    def get_beam_direction(self):
-        return 1
+
+def _get_ring_from_seqname(seq):
```
JoschD: still in use?

Issue #88, opened 2019-08-06
Overall accuracy test based on PTC


lmalina:
Add a test which tracks particles with PTC, runs frequency analysis, measures the optics and check the results.

mihofer:
test 3_BPM method (specifically alpha) as part of this tests, maybe on a small machine (PSB)

PR #172, opened 2020-02-10
157/bug/missing error of alpha

### LINKED ISSUE #157: calculation of error of alpha in three bpm is missing
currently column ERRALF{PLANE} is always set to 0, messes with accuracy test of opticmeasurements
Sidenote: Columnnames are still hardcoded, are not using constants.py

awegsche:
Fixes #157 

Error calculation for 3BPM alpha is added. Alpha calculation takes now all three combinations.

Tests extended to include 3BPM with adapted error tolerances. Tolerances to be discussed.


tests/accuracy/test_measure_optics.py
```diff
@@ -12,6 +12,7 @@
 from .twiss_to_lin import optics_measurement_test_files
 
 LIMITS = {'P': 1e-4, 'B': 3e-3, 'D': 1e-2, 'A': 6e-3}
+LIMITS_3BPM = {'P': 1e-4, 'B': 3e-3, 'D': 1e-2, 'A': 0.95}
```
lmalina: The limit for alpha function is not far off from situation, where LHC's arc are consistent with zero. I know the discrepancy comes mostly from IR, however pushes the limit on the test in this form too high. We should probably rethink the test of 3-BPM method ... what about test on PS lattice?

omc3/optics_measurements/beta_from_phase.py
```diff
@@ -419,21 +419,33 @@ def three_bpm_method(meas_input, phase, plane, meas_and_mdl_tunes):
...
+                + ((cot_phase_model + np.roll(cot_phase_model, -1, axis=0))[3] + 2.0 * alfmdl)
+                + ((cot_phase_model + np.roll(cot_phase_model, -2, axis=0))[1] + 2.0 * alfmdl)) / 6.0
+    alf_meas_term = (((cot_phase_meas + np.roll(cot_phase_meas, -1, axis=0))[0])
+                + ((cot_phase_meas + np.roll(cot_phase_meas, -1, axis=0))[3])
+                + ((cot_phase_meas + np.roll(cot_phase_meas, -2, axis=0))[1])) / 6.0
+
     # multiply the fractions by betmdl and calculate the arithmetic mean
     beti = bet_frac * betmdl
-    alfi = (bet_frac * (cot_phase_model[1] + cot_phase_model[3] + 2 * alfmdl) - (cot_phase_meas[1] + cot_phase_meas[3])) / 2
+    alfi = bet_frac * alf_mdl_term + alf_meas_term
```
lmalina: Should have just the average, or some weighting, for example, based on phase advances? This could avoid divergent behaviour.

tests/accuracy/test_measure_optics.py
mihofer: 3 BPM tests removed, will be implemented at a later point, likely as part of #88

PR #142, opened 2019-12-16
Enhancement/106/amplitude detuning

JoschD:

PR #182, opened 2020-02-28
Enhancement/175/update linfile

JoschD:
Merge after #142 as this is already based on the enhancement/106/amplitude_detuning branch
(should then also reduce the number of files to check... ??)


omc3/scripts/update_nattune_in_linfile.py
```diff
@@ -0,0 +1,189 @@
...
+            help=("List of paths to the spectrum files. The files need to be given"
+                  " without their '.lin'/'.amps[xy]','.freqs[xy]' endings. "
+                  " (So usually the path of the TbT-Data file.)")
+        ),
+        range=dict(
+            required=True,
+            nargs=2,
+            type=float,
+            help="Frequency range in which the highest peak should be found."
+        ),
```
lmalina: interval wound be a better name.

Issue #187, opened 2020-03-29
Complete example dataframe in phase.py


fsoubelet:
As we discussed a while ago with @lmalina, the phase dataframe given as example in the `calculate` function's docstring in the `omc3.optics_measurements.phase` module lacks a row, which is slightly misleading.

Since it can't hurt to fix it, I suggest simply adding the row.

PR #188, opened 2020-03-29
Refactor(docstring): #187 complete example matrix

fsoubelet:
The change is extremely simply.

Provided you're still ok with it, it should go through without breaking a sweat.


JoschD:
As I have no idea about that part of the code, I leave it to @lmalina

Issue #181, opened 2020-02-27
Create requirements.txt from setup.py 


JoschD:
Should have the options to create all, basic, tests and doc 
e.g. `python setup.py requirements all` 

hint: https://stackoverflow.com/questions/677577/distutils-how-to-pass-a-user-defined-parameter-to-setup-py

JoschD:
Any insights @Mael-Le-Garrec ?

PR #189, opened 2020-03-31
Requirement Printer

JoschD:
I had to make a new script for that, as adding commands in to setup.py, as envisioned in #181 was not possible (e.g. 
`setup.py requirements install` also runs install,  
`setup.py requirements --install` needs an argument for `--install`).

Please comment!

Usecase:

```
python requirements.py install > reqs.txt
python -m pip install -r reqs.txt
```

### DIFF
--- a/requirements.py
+++ b/requirements.py
@@ -1,0 +1,84 @@
+"""
+Requirement Printer
+"""
+import argparse
+from contextlib import suppress
+
+from setup import ALL_DEPENDENCIES
+
+KEYMAP = {'install': 'install_requires',
+          'test': 'tests_require',
+          'setup': 'setup_requires'}
+
+
+def print_requirements():
+    """
+    Prints the requirements for this package.
+
+    Use either 'install', 'test', 'setup' (if present in setup.py)
+    or one of the keys in 'extras_require'.
+    If no key is given ALL dependencies are printed.
+
+    """
+    keys = _parse_args()
+    _check_keys(keys)
+
+    if len(keys) == 0:
+        keys = _get_all_dependency_keys()
+
+    dependencies = []
+    for k in keys:
+        dependencies += _get_dependencies(k)
+    _print(dependencies)
+
+
+# Helper -----------------------------------------------------------------------
+
+
+def _parse_args():
+    parser = argparse.ArgumentParser(
+        description='Print arguments for this package'
+    )
+    parser.add_argument('keys',
+                        type=str, nargs='*', default=[],
+                        metavar="KEY",
+                        help="The requirements you want.")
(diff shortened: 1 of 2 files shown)


JoschD:
One of the bigger changes is that `setup.py` now has a `if __name__ == '__main__'`, which I couldn't find anywhere else and didn't know if it was allowed. But as travis seems to be fine with it, I am fine with it.

fsoubelet:
Is the goal to be able to install with, say, the testing requirements?

We could simply use the `extras` option from `setuptools`?
If we define a `tests` extra, then we can simply `pip install omc3[tests]` (for instance).

Or am I missing the point?

JoschD:
> Is the goal to be able to install with, say, the testing requirements? If so, we could simply use the [`extras`](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) option from `setuptools`.
> 
> For instance we can define a `tests` extra by adding `"tests": TEST_DEPENDENCIES` to the `EXTRA_DEPENDENCIES` variable. Then we can simply `pip install omc3[tests]`, or `pip install .[tests]` for now since we aren't deployed to PyPI.
> 
> Or am I missing the point?
> 
> EDIT: It already works with `doc` (installs sphinx et al) since we have it defined as an extra.
> omc3 doc

BUT if you do `pip install ".[doc]" it a) installs all 'main' requirements as well (not too bad) b) installs omc3 as well as a side-package. Which is what I was trying to avoid in the first place

fsoubelet:
If for development purposes we don't want to have `omc3` as a site-package, then there is pip's  `--editable` flag, which is basically setuptool's [`development mode`](https://setuptools.readthedocs.io/en/latest/setuptools.html#development-mode) with a few advantages, exactly for this purpose.

Then, to develop and test your workflow becomes (if we define a `tests` extra):
```bash
git clone https://github.com/pylhc/omc3
cd omc3
pip install --editable .[tests]
```

If you also want to build the docs, the last line becomes `pip install --editable .[tests, doc]`.

Tests (and doc) dependencies installed and you make your changes as usual without `omc3` being a site-package. Setuptool properly symlinked things so that you can run tests.

I think going the Python way and avoiding "obscure" code (for lack of a better word) is the thing to do. To me this approach feels simpler, more intuitive, and only requires a few lines in `setup.py` and a flag at install.

That's my opinion though. I'd also like @lmalina, @Mael-Le-Garrec, @mihofer, @awegsche and @jaimecp89's take.

JoschD:
And if one *really* does not want omc3 linked to the package one can just `pip uninstall omc3`
... should work

PR #190, opened 2020-04-01
#181 Easier dependency management

JoschD:
* Changed dependency layout
* added 'all'
* some pytest-args examples

Controversial:
* added pillow-requirement as old version kept crashing with matplotlib (https://github.com/matplotlib/matplotlib/issues/14263 )

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,17 +48,17 @@ with README.open("r") as docs:
 
 # Dependencies for the package itself
 DEPENDENCIES = [
-    "matplotlib==3.2.*",
-    "Pillow>=6.0.0",  # not our dependency but older versions crash with mpl
-    "numpy>=1.14.1",
+    "matplotlib>=3.2.0",
+    "Pillow>=6.2.2",  # not our dependency but older versions crash with mpl
+    "numpy>=1.18.0",
     "pandas==0.25.*",
-    "scipy>=1.0.0",
-    "scikit-learn>=0.20.3",
+    "scipy>=1.4.0",
+    "scikit-learn>=0.22.0",
     "tfs-pandas>=1.0.3",
     "generic-parser>=1.0.6",
     "sdds>=0.1.3",
     "pytz>=2018.9",
-    "h5py>=2.7.0",
+    "h5py>=2.9.0",
     "pytimber>=2.8.0",
 ]
 
@@ -69,8 +69,8 @@ EXTRA_DEPENDENCIES = {
     ],
     "test": [
         "pytest>=5.2",
-        "pytest-cov>=2.6",
-        "hypothesis>=3.23.0",
+        "pytest-cov>=2.7",
+        "hypothesis>=5.0.0",
         "attrs>=19.2.0",
     ],
     "doc": [
--- a/setup.py
+++ b/setup.py
@@ -48,8 +48,8 @@ with README.open("r") as docs:
 
 # Dependencies for the package itself
 DEPENDENCIES = [
-    "matplotlib>=3.1.0",
-    "pillow>=6.0.0",  # not our dependency but older versions crash with mpl
+    "matplotlib==3.2.*",
+    "Pillow>=6.0.0",  # not our dependency but older versions crash with mpl
     "numpy>=1.14.1",
     "pandas==0.25.*",
     "scipy>=1.0.0",
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
 import pathlib
+import shlex
 import sys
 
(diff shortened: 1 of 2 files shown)


JoschD:
I also removed the main-dependencies from the `tests_require` as they are installed anyway, so it was redundant.

setup.py
```diff
@@ -41,6 +49,7 @@ def run_tests(self):
 # Dependencies for the package itself
 DEPENDENCIES = [
     "matplotlib>=3.1.0",
+    "pillow>=6.0.0",  # not our dependency but older versions crash with mpl
```
fsoubelet, line 52: The actually package is `Pillow`, though I don't know if the lowercase P could be an issue. Can't hurt to make the change.

setup.py
```diff
@@ -53,19 +62,26 @@ def run_tests(self):
...
+        "attrs>=19.2.0",
+    ],
+    "doc": [
+        "sphinx",
+        "travis-sphinx",
+        "sphinx_rtd_theme"
+    ],
+}
+EXTRA_DEPENDENCIES.update(
+    {'all': [elem for list_ in EXTRA_DEPENDENCIES.values() for elem in list_]}
```
fsoubelet, line 83: Some verbosity can't hurt. Maybe `[requirement for extra_reqs in EXTRA_DEPENDENCIES.values() for requirement in extra_reqs]`?

JoschD:
From previous travis runs it looks like python 3.8 does not like mpl 3.1.0 in this combo. Shall we specify mpl == 3.2.* ?

setup.py
fsoubelet: @JoschD I just saw that matplotlib has a `Pillow>=6.2.0"` requirement, is this one needed then?

JoschD:
>to include the README changes

They are already merged

setup.py
JoschD: I don't think it makes it any more readable and just makes the line much longer.

fsoubelet:
> They are already merged

My bad.

lmalina:
> From previous travis runs it looks like python 3.8 does not like mpl 3.1.0 in this combo. Shall we specify mpl == 3.2.* ?
 I guess, we should specify matplotlib >=3.2.0, this is also a good PR to check if the other minimal versions are still valid.
I am certain about pandas. At some point I had troubles with numpy, so maybe numpy >=1.16.5, which was a solution, etc.

setup.py
JoschD: Weird. I have an old mpl and old Pillow (5.1.0) installed. 
Only the mpl requirement in the Dependencies.
Then I run `python setup.py install test`.
-> Installs mpl 3.2.1
crashes `ImportError: cannot import name '_imaging' from 'PIL' (/usr/lib/python3/dist-packages/PIL/__init__.py)`

With the explicit pillow dependency it updates pillow and runs fine.
Can you verify?

JoschD:
> I guess, we should specify matplotlib >=3.2.0, this is also a good PR to check if the other minimal versions are still valid.
@lmalina 
I think we should be careful with >=   ...  I'd stay on the current minor version, as higher versions might break things? Or go with >= and < and then change the limits after being tested

lmalina:
> > I guess, we should specify matplotlib >=3.2.0, this is also a good PR to check if the other minimal versions are still valid.
> 
> @lmalina
> I think we should be careful with >= ... I'd stay on the current minor version, as higher versions might break things? Or go with >= and < and then change the limits after being tested

I would not be too restrictive towards future versions. Of course, it is to be tested, but that is generally the case, for example, in travis. We find out quite soon if something breaks with newer version. In such cases, we can put upper limit, unless we fix it right away.

setup.py
fsoubelet: Indeed I can confirm, quite strange. Let's leave the Pillow requirement here.

JoschD:
> > > I guess, we should specify matplotlib >=3.2.0, this is also a good PR to check if the other minimal versions are still valid.
> > 
> > 
> > @lmalina
> > I think we should be careful with >= ... I'd stay on the current minor version, as higher versions might break things? Or go with >= and < and then change the limits after being tested
> 
> I would not be too restrictive towards future versions. Of course, it is to be tested, but that is generally the case, for example, in travis. We find out quite soon if something breaks with newer version. In such cases, we can put upper limit, unless we fix it right away.

I updated the lower versions. These are the ones I found (starting from what we had before) that will work on python 3.6-3.8 if you go with a vanilla python dist. and use exactly theses versions. Lower versions crashed for one reason or another.

Issue #198, opened 2020-05-26
Speed up calculations with pandas in Harpy


lmalina:
It much faster to do: 
```dataframe.iloc[:,:]=dataframe.iloc[:,:].to_numpy() * number``` 
than 
```dataframe = dataframe * number```

Issue #174, opened 2020-02-14
do (some) todos


mihofer:
accelerator kwargs in hole in one: write kwargs from accel class and/or point to accel class file and class files inheriting from this one, chromatic beating option etc. missing in docstring
updated readme: fix typo (clone which, perform frequency _analysis_ ), add packages tfs, sdds, generic-parser and links to repos
general: fix Future warnings

fsoubelet:
https://github.com/pylhc/omc3/blob/c3acd867dc105dbcb3a58ff15a29d5e44353b561/omc3/model/write_macros.py#L15

replace `:` with `;`

mihofer:
I think except for the future warnings all issues in here have been closed, so closure might be in order .. or keep it as a small issue bucket? Opinions @fsoubelet @JoschD @lmalina @awegsche

fsoubelet:
I'd say we close this issue once the branch in which fixes are implemented gets merged. I don't think keeping up an issue with a vague name is going to help in the long term compared to a few specific ones.

mihofer:
fixes were implemented in different branches, most of them afaik already merged to master, Issue is now linked with PR #191 as that's the last problem to be fixed

PR #199, opened 2020-05-26
Bug/197/svd cleaning

### LINKED ISSUE #197: Fix SVD cleaning
- subtract closed orbit (per BPM) before SVD calculation instead of mean of the whole matrix
- remove scaling by sqrt(N_turns) before SVD calculation
- make sure only a single BPM can be removed per mode per iteration
- add option for maximal number of iteration
- remove flag ``--keep_dominant_bpms``

lmalina:
Fixes #197 and #198. 
BPM cleaning based on SVD works now in up to a user-specified number of iteration, the modes are renormalised between iterations. 
Only one element can be removed per mode per iterations, consequently, only one BPM can be removed
Closed orbit is now subtracted before SVD and unnecessary scaling has been removed.
Calculations with pandas slowed down at some point (I assume by some change of indexing), now fixed. Harpy runs again around 3 seconds instead of 10 seconds.
Improves a few things pointed out in #174.


mihofer:
Not part of this issue but shouldn't _clean() always return only bpms present in the model? As it is done now, only if opt.clean is True, a selection based on the bpms is done.. useful ie. when loading trackone data which usually contains start and endpoint which are not in the model and should thus be removed  .. https://github.com/pylhc/omc3/blob/99b61e618a52874c1c9c4e870e092e9f409d473a/omc3/harpy/clean.py#L33 (..unfortunately github doesn't allow to comment on unchanged code .. )

omc3/harpy/clean.py
```diff
@@ -173,67 +174,55 @@ def _resync_bpms(harpy_input, bpm_data, model):
...
+            np.dot(np.diag(s_mat[indices]), vt_mat[indices, :]),
+            u_mat_mask[:, :int(np.max(indices))+1])
 
 
-def _remove_dominant_elements(u_mat, s_mat, dominance_limit):
+def _remove_dominant_elements(u_mat, s_mat, dominance_limit, num_iter=1):
     u_mat_mask = np.ones(u_mat.shape, dtype=bool)
     if dominance_limit is None:
         return u_mat, s_mat, u_mat_mask
     abs_dominance_limit = np.abs(dominance_limit)
```
mihofer, line 212: since signed_limit is not used anymore, this line would be redundant if options is checked in entrypoint

omc3/harpy/clean.py
```diff
@@ -173,67 +174,55 @@ def _resync_bpms(harpy_input, bpm_data, model):
...
-            u_mat_mask[:, indices])
+
+    indices = np.argsort(s_mat)[::-1][:keep]
+    return (pd.DataFrame(index=matrix.index, data=u_mat[:, indices]),
+            np.dot(np.diag(s_mat[indices]), vt_mat[indices, :]),
+            u_mat_mask[:, :int(np.max(indices))+1])
 
 
-def _remove_dominant_elements(u_mat, s_mat, dominance_limit):
+def _remove_dominant_elements(u_mat, s_mat, dominance_limit, num_iter=1):
```
JacquelineKeintzel, line 208: Here the default value of num_iter is '1', whereas in line 177 it is 'None' in the function svd_decomposition(). Is it on purpose?

lmalina: None is on purpose, 1 should be 3...fixed now

omc3/harpy/clean.py
```diff
@@ -31,9 +30,9 @@ def clean(harpy_input, bpm_data, model):
     Returns:
         Clean BPM matrix, its decomposition, bad BPMs summary and estimated BPM resolutions
     """
-    if not harpy_input.clean:
-        return bpm_data, None, [], None
     bpm_data, bpms_not_in_model = _get_only_model_bpms(bpm_data, model)
+    if not harpy_input.clean:
+        return bpm_data, None, bpms_not_in_model, None
     if bpm_data.empty:
```
mihofer, line 36: shouldn't this be also moved to the _get_only_model_bpms() i.e before the if return part?

PR #201, opened 2020-06-09
# Fixes #200:

### LINKED ISSUE #200: Parameters of beta_from_phase
ZERO_THRESHOLD and RCOND parameters are wrong in the code. It causes troubles:

rcond_beta

Negative betas and betas with error actually larger than the value should be explicitly filtered out.

lmalina:
corrected parameter values
    explicitly filtered out negative betas and too high errors

### DIFF
--- a/omc3/optics_measurements/beta_from_phase.py
+++ b/omc3/optics_measurements/beta_from_phase.py
@@ -25,9 +25,9 @@ from omc3.utils import logging_tools, stats
 LOGGER = logging_tools.get_logger(__name__)
 
 EPSILON = 1.0E-16
-ZERO_THRESHOLD = 1e-3
+ZERO_THRESHOLD = 1e-2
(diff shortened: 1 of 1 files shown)


omc3/optics_measurements/beta_from_phase.py
```diff
@@ -162,6 +162,10 @@ def n_bpm_method(meas_input, phase, plane, meas_and_mdl_tunes):
     beta_df["NCOMB"] = n_comb
     LOGGER.debug(f"No valid combinations for BPMs: {list(beta_df.index[beta_df['NCOMB'] == 0])}.")
     beta_df = beta_df.loc[beta_df["NCOMB"] > 0]
+    beta_df = beta_df.loc[beta_df[f"BET{plane}"] > 0]
```
mihofer, line 165: maybe log somewhere the BPM with negative beta fct

PR #211, opened 2020-09-03
FIxes #210: aligned longitudinal phase to correct reference turn

### LINKED ISSUE #210: Harpy: Align longitudinal phase
Phases of all spectral lines are aligned to the middle of analysed turns, to limit phase error originating from error in frequency. For example, if 6600 turns are analysed the resulting phases of spectral lines are valid in turn number 3300. However, the phase of synchrotron line "MUZ" is not aligned with the rest, i.e. the reference turn number is 0.

lmalina:
also changing some .values to .to_numpy()

### DIFF
--- a/tests/accuracy/test_harpy.py
+++ b/tests/accuracy/test_harpy.py
@@ -18,7 +18,10 @@ COUPLING = 0.01
 NTURNS = 1024
 NBPMS = 100
 BASEAMP = 0.001
+AMPZ, MUZ, TUNEZ = 0.01, 0.3, 0.008
+
 np.random.seed(1234567)
+
 
 class BasicTests:
     @staticmethod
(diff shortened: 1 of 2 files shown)


tests/accuracy/test_harpy.py
```diff
@@ -124,14 +149,17 @@ def _get_model_dataframe():
...
-                     model.loc[:, "TUNEX"].values[:, None] * ints[None, :]))
-    data_y = model.loc[:, "AMPY"].values[:, None] * np.cos(
-        2 * np.pi * (model.loc[:, "MUY"].values[:, None] +
-                     model.loc[:, "TUNEY"].values[:, None] * ints[None, :]))
+    data_x = model.loc[:, "AMPX"].to_numpy()[:, None] * np.cos(
+        2 * np.pi * (model.loc[:, "MUX"].to_numpy()[:, None] +
+                     model.loc[:, "TUNEX"].to_numpy()[:, None] * ints[None, :]))
+    data_y = model.loc[:, "AMPY"].to_numpy()[:, None] * np.cos(
+        2 * np.pi * (model.loc[:, "MUY"].to_numpy()[:, None] +
+                     model.loc[:, "TUNEY"].to_numpy()[:, None] * ints[None, :]))
```
mihofer, line 157: wouldn't that be something for {plane: formulas for plane in PLANES} ?

tests/accuracy/test_harpy.py
```diff
@@ -124,14 +149,17 @@ def _get_model_dataframe():
...
+    data_z = AMPZ * BASEAMP * np.ones((NBPMS, 1)) * np.cos(
+        2 * np.pi * (MUZ * np.ones((NBPMS, 1)) +
+                     TUNEZ * np.ones((NBPMS, 1)) * ints[None, :]))
     mats = dict(X=pd.DataFrame(data=np.random.randn(model.index.size, NTURNS) * NOISE + data_x
-                               + COUPLING * data_y, index=model.index),
+                               + COUPLING * data_y + data_z, index=model.index),
                 Y=pd.DataFrame(data=np.random.randn(model.index.size, NTURNS) * NOISE + data_y
                                + COUPLING * data_x, index=model.index))
```
mihofer, line 164: why is data_z only added in the X case?

lmalina: that relates to dispersion, which we often have only in horizontal plane...for the tests it does not really matter

mihofer: then probably a DISP factor similar to COUPLING would make it easier understandable

tests/accuracy/test_harpy.py
lmalina: that is not really part of this PR...just changed .values to .to_numpy() here. I added the data_z, which would on contrary break the possible convention.

mihofer (approved):

tests/accuracy/test_harpy.py
lmalina: there is an AMPZ factor...as the factor should physically relate to D_x * dp/p

awegsche (approved):

PR #217, opened 2020-09-08
Enhancement/164/coupled action

lmalina:
Introduces correction to coupling in action calculation. For the moment waits for coupling module to be finished and lacks a test.

### DIFF
--- a/omc3/optics_measurements/kick.py
+++ b/omc3/optics_measurements/kick.py
@@ -77,14 +77,22 @@ def _getkick(measure_input, files, plane):
     return kick_frame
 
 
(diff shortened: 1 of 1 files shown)


omc3/optics_measurements/kick.py
```diff
@@ -78,14 +78,26 @@ def _getkick(measure_input, files, plane):
     return kick_frame
 
 
-def _gen_kick_calc(meas_input, lin, plane):
+def _gen_kick_calc(meas_input, lin, plane, coupling=None):
```
mihofer, line 81: add header in argument to put a flag there if correction was applied?

PR #221, opened 2020-09-16
Bug/219/madx macros and more

mihofer:
Moving lib content to model/madx_macros
Include non .py files when installing with `pip install git+https://github.com/pylhc/omc3.git`

Change test handling

Remove python 3.6 from tests and requirements, after merge only 3.7 and above is maintained, corrdinated merge with new tfs


setup.py
```diff
@@ -6,33 +6,6 @@
 from setuptools.command.test import test as TestCommand
```
fsoubelet, line 6: Shouldn't be needed anymore

mihofer: fixed

fsoubelet:
Do we need a `pytest.ini` just to describe two markers?

mihofer:
> 
> 
> Do we need a `pytest.ini` just to describe two markers?

Kinda yes, kinda no .. if the markers are not specified, it will complain every time that it could be wrong and running it with the --strict-markers would fail then ( https://docs.pytest.org/en/stable/mark.html ) .. one could also define it somewhere programmatically but pytest.ini looked more convenient to me

fsoubelet:
Fair enough. Wouldn't we want to group most tests in verbosely named classes and `mark` the classes to avoid getting decorators everywhere? Grouping test functions in classes around common themes would also let us easily test a single thematic with the `-k` option from `pytest`.

Additionally, does calling `pytest -m basic` also run functions that aren't decorated? If not, maybe we should change the CI command to `pytest -m "not basic"` in `.travis.yml` to ensure that branch builds run everything except the extended tests.

fsoubelet (changes requested):
Can you remove `pytest-runner` and `setup_requires` as discussed?

Seems to me the `shlex` and `sys` imports aren't needed anymore in `setup.py`.

mihofer:
> 
> 
> Can you remove `pytest-runner` and `setup_requires` as discussed?
> 
> Seems to me the `shlex` and `sys` imports aren't needed anymore in `setup.py`.

`pytest-runner` remains should be gone now and unused imports removed

fsoubelet (dismissed):

fsoubelet:
Seems all good to me. Should rerun builds when `tfs-pandas` is ready for `2.0` to check that all works fine, and we'll be good.

fsoubelet:
Actually I don't think we need `hypothesis` and `attrs` as test dependencies either @mihofer

lmalina:
> Actually I don't think we need `hypothesis` and `attrs` as test dependencies either @mihofer

`attrs` is not our direct dependency, however tests in travis wouldn't run without it

fsoubelet:
@mihofer do we need all the `@pytest.mark.basic` markers actually? Considering that for "basic" tests we run `-m 'not extended'`

mihofer:
> 
> 
> @mihofer do we need all the `@pytest.mark.basic` markers actually? Considering that for "basic" tests we run `-m 'not extended'`

as it is now no, though I think for completeness it's not bad to mark every test .. seeing issue #228 we could then also add a `fast` marker to certain tests

setup.cfg
```diff
@@ -3,7 +3,9 @@ codestyle_max_line_length = 100
 codestyle_exclude = doc/conf.py
 addopts = --cov-report xml --cov=omc3
 testpaths = tests/unit tests/accuracy
-python_classes = BasicTests ExtendedTests
+markers=
+        basic: basic tests run for every commit
+        extended: test run on PRs
```
lmalina, line 8: just one of them should be enough

setup.py
```diff
@@ -50,11 +20,11 @@ def run_tests(self):
 DEPENDENCIES = [
     "matplotlib>=3.2.0",
     "Pillow>=6.2.2",  # not our dependency but older versions crash with mpl
-    "numpy>=1.18.0",
-    "pandas==0.25.*",
-    "scipy>=1.4.0",
+    "numpy>=1.19.0",
+    "pandas>=1.0",
+    "scipy>=1.5.0",
     "scikit-learn>=0.22.0",
```
lmalina, line 26: scikit-learn>=0.23.0

lmalina (changes requested):

setup.py
mihofer: more information plz? i hope you didnt type it on your smartphone so a bit more writing can be expected ...

setup.cfg
mihofer: should be enough but there is also no solid reason why to scrape it, is there? and anybody in the future might appreciate to find a reason why we distinguish

omc3/madx_wrapper.py
```diff
@@ -22,7 +22,7 @@
 
 LOG = logging_tools.get_logger(__name__)
 
-LIB = abspath(join(dirname(__file__), "lib"))
+LIB = abspath(join(dirname(__file__), "model", "madx_macros"))
```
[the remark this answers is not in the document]
mihofer: moved to tests

setup.py
lmalina: first version compatible with python 3.8

lmalina (approved):

setup.cfg
```diff
@@ -3,7 +3,9 @@ codestyle_max_line_length = 100
 codestyle_exclude = doc/conf.py
```
fsoubelet, line 3: Delete lines 2 and 3 since they're meant for `pycodestyle`

setup.py
```diff
@@ -50,11 +20,11 @@ def run_tests(self):
 DEPENDENCIES = [
     "matplotlib>=3.2.0",
     "Pillow>=6.2.2",  # not our dependency but older versions crash with mpl
-    "numpy>=1.18.0",
-    "pandas==0.25.*",
-    "scipy>=1.4.0",
-    "scikit-learn>=0.22.0",
-    "tfs-pandas>=1.0.3",
+    "numpy>=1.19.0",
+    "pandas>=1.0",
```
fsoubelet, line 24: Do really we need to force pandas `>=1.0`?

Leaving `>=0.25` instead will let `pip` install `1.x` which we're compatible with but give some more freedom in environments.

fsoubelet (dismissed):

setup.py
mihofer: But new tfs requirements are also >=1.0 so  i think there is nothing to be gained there as well, is there?

fsoubelet: I guess you're right. @lmalina do you approve?

lmalina: we need pandas >=1.0 because of .convert_dtypes()

lmalina (approved):

fsoubelet (approved):

fsoubelet:
Good to me 👍

Issue #223, opened 2020-09-21
add Lumi Imbalance script


mihofer:
add script /afs/cern.ch/eng/sl/lintrack/LHC_commissioning2017/LumiImbalance.py to calc lumi imbalance from kmod results into omc3

Mael-Le-Garrec:
Where would you put it in omc3? I wanna do it :)

mihofer:
omc3/omc3/scripts .. my idea of an implementation was to read a twiss file where the kmod results are compiled (i.e rows IP1B1, IP1B2, IP5B1, IP5B2) and then perform the calculation of the imbalance and print to cmd/ add to header of the input file .. 
(and dont forget to add to README)

lmalina:
This should be probably combined with a script to load the kmod results per IP and per beam

PR #225, opened 2020-09-22
Script for luminosity imbalance

Mael-Le-Garrec:
Fixes https://github.com/pylhc/omc3/issues/223


omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,118 @@
...
+    expected_columns = ('BETSTARX', 'BETSTARY', 'ERRBETSTARX', 'ERRBETSTARY')
+    if not all([column in df.columns for column in expected_columns]):
+        LOG.error("Expected columns in the TFS file not found. Expected "
+                  "columns: {}".format(expected_columns))
+        raise KeyError('Columns not found in TFS file')
+
+
+def get_imbalance(df):
+    # Get the individual rows for the IP1 and IP5
+    ip1 = df.loc[df['LABEL'].str.startswith('IP1')]
```
JoschD, line 47: Loop over ips? -> half the code

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,118 @@
...
+def _validate_tfs(df):
+    # First validate the labels, which correspond to the IP and Beam
+    expected_labels = set(('IP1B1', 'IP1B2', 'IP5B1', 'IP5B2'))
+    if set(df['LABEL']) != expected_labels:
+        LOG.error("Expected labels in the TFS file not found. Expected labels:"
+                  " {}".format(expected_labels))
+        raise KeyError('Labels not found in TFS file')
+
+    # Validate the columns we need: BETSTAR{X,Y} and ERRBETSTAR{X,Y}
+    expected_columns = ('BETSTARX', 'BETSTARY', 'ERRBETSTARX', 'ERRBETSTARY')
```
JoschD, line 38: Column-Names as constants + plane (see rest of omc3)

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,118 @@
...
+            'relative_error': relative_error,
+            'eff_beta_ip1': eff_beta_ip1, 
+            'rel_error_ip1': rel_error_ip1,
+            'eff_beta_ip5': eff_beta_ip5, 
+            'rel_error_ip5': rel_error_ip5
+            }
+
+
+def print_luminosity(res):
+    print('{:22s}'.format("Luminosity imbalance:"), res['imbalance'])
```
JoschD, line 86: LOG.info()

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,118 @@
+from math import sqrt
+
+import tfs
+
+from omc3.utils.logging_tools import get_logger, list2str
+from generic_parser import entrypoint, EntryPointParameters
+from generic_parser.entrypoint_parser import save_options_to_config
```
mihofer, line 7: list2str is not used, same for save_options?

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,118 @@
...
+    print('{:22s}'.format("Relative error:"), res['relative_error'])
+    print('{:22s}'.format("Effective beta IP1:"), res['eff_beta_ip1'])
+    print('{:22s}'.format("Rel error IP1:"), res['rel_error_ip1'])
+    print('{:22s}'.format("Effective beta IP5:"), res['eff_beta_ip5'])
+    print('{:22s}'.format("Rel error IP5:"), res['rel_error_ip5'])
+
+
+@entrypoint(get_params(), strict=True)
+def main(opt):
+    df = tfs.read_tfs(opt['tfs'])
```
JoschD, line 96: opt is DotDict, so you can/could use opt.tfs, opt.inplace etc.

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+
+def print_luminosity(res):
+    print('{:22s}'.format("Luminosity imbalance:"), res['imbalance'])
+    print('{:22s}'.format("Relative error:"), res['relative_error'])
+    print('{:22s}'.format("Effective beta IP1:"), res['eff_beta_ip1'])
+    print('{:22s}'.format("Rel error IP1:"), res['rel_error_ip1'])
+    print('{:22s}'.format("Effective beta IP5:"), res['eff_beta_ip5'])
+    print('{:22s}'.format("Rel error IP5:"), res['rel_error_ip5'])
```
mihofer, line 91: Lukas will complain that it's not f strings .. 
also maybe using tfs.tools. significant digits could be used to shorten the output

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,118 @@
...
+                help="Add the luminosity imbalance in the source TFS file as a"
+                     "header"
+            )
+    )
+
+
+def _validate_tfs(df):
+    # First validate the labels, which correspond to the IP and Beam
+    expected_labels = set(('IP1B1', 'IP1B2', 'IP5B1', 'IP5B2'))
+    if set(df['LABEL']) != expected_labels:
```
JoschD, line 32: are you allowed to have labels twice?
would check here OR load tfs with index='LABEL'

tests/unit/test_imbalance.py
```diff
@@ -0,0 +1,91 @@
+import os
+import shutil
+import tempfile
+from datetime import datetime
+
+import h5py
```
mihofer, line 6: h5py is also not needed in the test?

JoschD (changes requested):

mihofer (dismissed):
Small things but good to go from my side

omc3/scripts/luminosity_imbalance.py
Mael-Le-Garrec: I didn't get the part about significant digits, what's the desired output?

omc3/scripts/luminosity_imbalance.py
Mael-Le-Garrec: I prefer improved readability over number of lines

omc3/scripts/luminosity_imbalance.py
mihofer: so instead of having an output of 0.123456789 +- 0.00123456 , printing just 0.123 +- 0.001 would be more sensible .. since we only measure with the precision of the error, any digits in the actual result would not be particularly meaningful
https://en.wikipedia.org/wiki/Significant_figures

omc3/scripts/luminosity_imbalance.py
Mael-Le-Garrec: where would you put them?

omc3/scripts/luminosity_imbalance.py
Mael-Le-Garrec: Why?

omc3/scripts/luminosity_imbalance.py
JoschD: ok, then:

Loop over ips? -> more readable code as less line repetitions

omc3/scripts/luminosity_imbalance.py
JoschD: because we don't print.

all outputs are logged, so you have control. Also, if you use logging and print in the same script you might get weird line-mixes as logging runs in parallel to the code

omc3/scripts/luminosity_imbalance.py
Mael-Le-Garrec: I understand what significant digits are but why would you want to shorten the output?

JoschD: readability

tests/unit/test_imbalance.py
```diff
@@ -0,0 +1,86 @@
...
+
+import tfs
+
+from omc3.scripts import luminosity_imbalance
+
+CURRENT_DIR = os.path.dirname(__file__)
+
+
+def test_result_tfs(_tfs_file):
+    res = luminosity_imbalance.main({'tfs': _tfs_file})
```
JoschD, line 14: also hint:
`res = luminosity_imbalance.main(tfs=_tfs_file)`

looks more function-like?

tests/unit/test_imbalance.py
```diff
@@ -0,0 +1,86 @@
...
+    with pytest.raises(KeyError) as error:
+        luminosity_imbalance.main({'tfs': _tfs_wrong_label})
+    assert 'Labels not found in TFS file' in str(error.value)
+
+
+def _get_file(tmp_path, path):
+    # Copy the file to a temp directory so that we don't modify the source
+    d = tmp_path / 'imbalance'
+    d.mkdir()
+    src = os.path.join(CURRENT_DIR, os.pardir, "inputs", "lumi_imbalance",
```
JoschD, line 64: pathlib as above

tests/unit/test_imbalance.py
```diff
@@ -0,0 +1,86 @@
+import os
+import shutil
+
+import pytest
+
+import tfs
+
+from omc3.scripts import luminosity_imbalance
+
+CURRENT_DIR = os.path.dirname(__file__)
```
JoschD, line 10: pathlib

omc3/scripts/luminosity_imbalance.py
JoschD: You can get PLANES from https://github.com/pylhc/omc3/blob/master/omc3/definitions/constants.py

and I'd put ERR and BETASTAR at the top of this script

omc3/scripts/load_kmod.py
```diff
@@ -0,0 +1,60 @@
...
+
+def loader_params():
+    params = EntryPointParameters()
+    params.add_parameter(name="kmod_dir", type=str, required=True,
+                         help="path to kmod directory with stored KMOD measurement files")
+    params.add_parameter(name="res_dir", type=str, required=True, help="Output directory of the optics analysis.")
+    params.add_parameter(name="mod_path", type=str, required=True,
+                         help="Specify path to current model.")
+    params.add_parameter(name="realizations", type=str, choices=("B1", "B2"),
+                         help="Number of copies with added noise")
```
mihofer, line 26: ?? .. also is it used?

tests/unit/test_imbalance.py
Mael-Le-Garrec: I just followed what has already been done. I'm confused, what should I do???? https://github.com/pylhc/omc3/blob/master/tests/unit/test_tbt.py#L14

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,144 @@
...
+                help="Add the luminosity imbalance in the source TFS file as a"
+                     "header"
+            )
+    )
+
+
+def _validate_tfs(df):
+    # First validate the labels, which correspond to the IP and Beam
+    expected_labels = set(LABELS)
+    if len(df[LABEL]) != 4 or set(df[LABEL]) != expected_labels:
```
JoschD, line 42: `!= len(LABELS)`

tests/unit/test_imbalance.py
Mael-Le-Garrec: I prefer the dict, it conveys more the idea that there is the entrypoint behind

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,144 @@
...
+            'relative_error': relative_error,
+            'eff_beta_ip1': eff_beta[IPS[0]],
+            'rel_error_ip1': rel_error[IPS[0]],
+            'eff_beta_ip5': eff_beta[IPS[1]],
+            'rel_error_ip5': rel_error[IPS[1]]
+            }
+
+
+def print_luminosity(res):
+    LOG.info('{:22s}: {}'.format("Luminosity imbalance", res['imbalance']))
```
JoschD, line 104: what happened to f-strings? ^^'

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,144 @@
...
+        error = {}
+        for p in PLANES:
+            average_beta[p] = row.sum(axis=0)[f'{BETASTAR}{p}'] / 2
+            error[p] = 0.5 * sqrt((row[f'{ERR}{BETASTAR}{p}'] ** 2).sum(axis=0))
+
+        # Compute the relative error for this IP
+        rel_error[ip] = 0
+        for p in PLANES:
+            rel_error[ip] += (error[p] ** 2) / (average_beta[p] ** 2)
+        rel_error[ip] = 0.5 * sqrt(rel_error[ip])
```
JoschD, line 83: `rel_error[ip] = 0.5 * sqrt(sum([(error[p] ** 2) / (average_beta[p] ** 2) for p in planes]))`

if you like

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,144 @@
...
+
+        # Get the effective beta
+        eff_beta[ip] = 1 / sqrt(1 / (reduce(mul, average_beta.values())))
+
+    # Compute the whole relative error
+    relative_error = sum(rel_error.values())
```
JoschD, line 89: shouldn't that also be the RMS? (asking @mihofer and @lmalina )

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,144 @@
...
+    for ip in IPS:
+        # Get the individual rows for the IP1 and IP5
+        row = df.loc[df[LABEL].str.startswith(ip)]
+
+        # Compute the average β* for each plane with both beams
+        # and the absolute errors for each plane
+        average_beta = {}
+        error = {}
+        for p in PLANES:
+            average_beta[p] = row.sum(axis=0)[f'{BETASTAR}{p}'] / 2
```
JoschD, line 76: `row[f'{BETASTAR}{p}'].sum(axis=0)`
So it doesn't calculate the sum of all columns

omc3/scripts/luminosity_imbalance.py
Mael-Le-Garrec: true

omc3/scripts/luminosity_imbalance.py
Mael-Le-Garrec: they were never here in the first place, they're in Urlaub

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,144 @@
...
+            error[p] = 0.5 * sqrt((row[f'{ERR}{BETASTAR}{p}'] ** 2).sum(axis=0))
+
+        # Compute the relative error for this IP
+        rel_error[ip] = 0
+        for p in PLANES:
+            rel_error[ip] += (error[p] ** 2) / (average_beta[p] ** 2)
+        rel_error[ip] = 0.5 * sqrt(rel_error[ip])
+
+        # Get the effective beta
+        eff_beta[ip] = 1 / sqrt(1 / (reduce(mul, average_beta.values())))
```
JoschD, line 86: why not `np.prod()` ?

tests/unit/test_imbalance.py
JoschD: We are trying to use pathlib from now on everywhere. 
Also temp_path is already a `Path` so it would be much more in line. 

Seems that `test_tbt.py` could use some pathlib update

tests/unit/test_imbalance.py
JoschD: But the whole idea of the entrypoint is, that it is usable from python as a function :D 
That's why we have the entrypoint-wrapper

lmalina:
I see the discussion blows up a bit. Why not start first with a merge of the two scripts? The situation will be clearer, no need to check some things etc.

omc3/scripts/luminosity_imbalance.py
Mael-Le-Garrec: because I didn't want to import numpy

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,142 @@
...
+                help="Add the luminosity imbalance in the source TFS file as a"
+                     "header"
+            )
+    )
+
+
+def _validate_tfs(df):
+    # First validate the labels, which correspond to the IP and Beam
+    expected_labels = set(LABELS)
+    if len(df[LABEL]) != len(LABELS) or set(df[LABEL]) != expected_labels:
```
JoschD, line 42: Actually, now that I think about it... you should allow other labels as well, as you are filtering by IP anyway. In case k-mod data includes other IPs. So your check would be like below:
```
not_found_labels = [l for l in expected_labels if l not in df[LABEL]]
if any(not_found_labels):
    msg = f'The following required labels are not found in dataframe: {str(not_found_labels).strip("[]")}`
```

omc3/scripts/luminosity_imbalance.py
JoschD: numpy is imported somewhere down the line anyway (tfs, pandas)

omc3/scripts/luminosity_imbalance.py
JoschD: `list2str` from before can be used to shorten

tests/unit/test_imbalance.py
```diff
@@ -0,0 +1,116 @@
...
+from pathlib import Path
+import pytest
+
+import tfs
+
+from omc3.definitions.constants import PLANES
+from omc3.scripts import luminosity_imbalance
+from omc3.scripts.luminosity_imbalance import BETASTAR, ERR
+
+CURRENT_DIR = Path(__file__)
```
JoschD, line 13: Sorry to be pendantic, but CURRENT_DIR would be `Path(__file__).parent`

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,166 @@
...
+    new_data = merge_tfs(ip_dir_names, 'result.tfs')
+    
+    # Get the imbalance
+    _validate_tfs(new_data)
+    res = get_imbalance(new_data)
+    res = get_significant_digits(res)
+    print_luminosity(res)
+
+    # Combine the lsa data
+    lsa_tfs = merge_tfs(ip_dir_names, 'lsa_reuslts.tfs')
```
mihofer, line 151: http://gph.is/19Wq15P typo alert typo altert

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,175 @@
...
+@entrypoint(loader_params(), strict=True)
+def merge_and_copy_kmod_output(opt):
+    # Get the directories we need where the tfs are stored
+    ip_dir_names = get_ip_dir_names(opt.kmod_dirs)
+
+    # Combine the data into one tfs
+    new_data = merge_tfs(ip_dir_names, 'results.tfs')
+
+    # Get the imbalance
+    _validate_tfs(new_data)
```
JoschD, line 154: So if I see this correctly this merger can only merge if there are IP1 XY and IP5 XY data present?

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,175 @@
...
+    return params
+
+
+@entrypoint(loader_params(), strict=True)
+def merge_and_copy_kmod_output(opt):
+    # Get the directories we need where the tfs are stored
+    ip_dir_names = get_ip_dir_names(opt.kmod_dirs)
+
+    # Combine the data into one tfs
+    new_data = merge_tfs(ip_dir_names, 'results.tfs')
```
JoschD, line 151: For easy access I'd make the tfs-files constants at the top of the file

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,175 @@
...
+
+    # Get the imbalance
+    _validate_tfs(new_data)
+    res = get_imbalance(new_data)
+    res = get_significant_digits(res)
+    print_luminosity(res)
+
+    # Combine the lsa data
+    lsa_tfs = merge_tfs(ip_dir_names, 'lsa_results.tfs')
```
JoschD, line 160: For easy access I'd make the tfs-files constants at the top of the file

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,175 @@
...
+    lsa_tfs.headers.update({"LUMINOSITY_IMBALANCE": res['imbalance'],
+                            "RELATIVE_ERROR": res['relative_error'],
+                            "EFF_BETA_IP1": res['eff_beta_ip1'],
+                            "REL_ERROR_IP1": res['rel_error_ip1'],
+                            "EFF_BETA_IP5": res['eff_beta_ip5'],
+                            "REL_ERROR_IP5": res['rel_error_ip5']
+                            })
+
+    # and write the resulting tfs
+    tfs.write(opt.res_dir / 'lsa_results.tfs', lsa_tfs)
```
JoschD, line 171: For easy access I'd make the tfs-files constants at the top of the file

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,175 @@
...
+                              "measurement files")
+    params.add_parameter(name="res_dir", type=pathlib.Path, required=True,
+                         help="Output directory where to write the result tfs")
+    return params
+
+
+@entrypoint(loader_params(), strict=True)
+def merge_and_copy_kmod_output(opt):
+    # Get the directories we need where the tfs are stored
+    ip_dir_names = get_ip_dir_names(opt.kmod_dirs)
```
JoschD, line 148: I'd have another input argument to set the data-labels manually (so you don't have to rename the folders) and only if that is not given to call this function

JoschD (changes requested):

lmalina:
Following from Eq(17) in [Concept of Luminosity](https://cds.cern.ch/record/941318/files/p361.pdf), the Luminosity imbalance is given by:

\frac{L_{IP1}}{L_{IP5}}=\frac{\sqrt{\beta_{x1,IP5}+\beta_{x2,IP5}}\cdot\sqrt{\beta_{y1,IP5}+\beta_{y2,IP5}}}{\sqrt{\beta_{x1,IP1}+\beta_{x2,IP1}}\cdot\sqrt{\beta_{y1,IP1}+\beta_{y2,IP1}}}

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def _validate_for_imbalance(df):
+    # First validate the labels, which correspond to the IP and Beam
+    expected_labels = set(LABELS)
+    not_found_labels = [label for label in expected_labels
+                        if not df[LABEL].str.contains(label).any()]
+
+    if any(not_found_labels):
+        #msg = 'The following required labels are not found in dataframe:' \
```
fsoubelet, line 35: No logging?

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,187 @@
...
+        if len(df.loc[df[LABEL] == label]) > 1:
+            msg = f'Found label {label} several times. Expected only once.'
+            LOG.error(msg)
+            raise KeyError(msg)
+
+    # Validate the columns we need: BETSTAR{X,Y} and ERRBETSTAR{X,Y}
+    expected_columns = [f'{BETASTAR}{p}' for p in PLANES] + \
+                       [f'{ERR}{BETASTAR}{p}' for p in PLANES]
+    if not all([column in df.columns for column in expected_columns]):
+        #msg = 'Expected columns in the TFS file not found. Expected ' \
```
fsoubelet, line 52: Same

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,187 @@
...
+    res['imbalance'], res['relative_error'] = significant_digits(
+            res['imbalance'], res['relative_error'])
+    res['eff_beta_ip1'], res['rel_error_ip1'] = significant_digits(
+            res['eff_beta_ip1'], res['rel_error_ip1'])
+    res['eff_beta_ip5'], res['rel_error_ip5'] = significant_digits(
+            res['eff_beta_ip5'], res['rel_error_ip5'])
+    return res
+
+
+def merge_tfs(directories, filename):
```
fsoubelet, line 117: You will be losing the headers if you `.append()` TfsDataFrames. Are we sure this is not a problem here?

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,187 @@
...
+
+        new_tfs = new_tfs.append(data, ignore_index=True)
+
+    return new_tfs
+
+
+def get_ip_dir_names(kmod_dirs):
+    # Check directories first
+    for d in kmod_dirs:
+        print(d)
```
fsoubelet, line 132: Should go

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,187 @@
...
+        if not d.exists():
```
fsoubelet, line 133: Use `.is_dir()` instead, `.exists()` will return `True` for a file

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,187 @@
...
+    return ip_dir_names
+
+
+def loader_params():
+    params = EntryPointParameters()
+    params.add_parameter(name="kmod_dirs", type=pathlib.Path,
+                         nargs='+', required=True,
+                         help="Path to kmod directories with stored KMOD "
+                              "measurement files")
+    params.add_parameter(name="res_dir", type=pathlib.Path, required=True,
```
fsoubelet, line 152: Shouldn't this be `outputdir` for consistency across `omc3` scripts?

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,187 @@
...
+
+BETASTAR = 'BETSTAR'
+ERR = 'ERR'
+LABEL = 'LABEL'  # Column containing the IP/Beam names
+BEAMS = ('B1', 'B2')
+IPS = ('ip1', 'ip5')
+LABELS = [f'{i}{b}' for i in IPS for b in BEAMS]
+
+
+def _validate_for_imbalance(df):
```
fsoubelet, line 28: Validation should rather not return anything and simply raise in case it detects an issue so we save indentation levels in scripts. @JoschD @lmalina ?

omc3/scripts/luminosity_imbalance.py
```diff
@@ -0,0 +1,187 @@
...
+    ip_dir_names = get_ip_dir_names(opt.kmod_dirs)
+
+    # Combine the data into one tfs
+    new_data = merge_tfs(ip_dir_names, f'{RESULTS}{EXT}')
+
+    # Combine the lsa data
+    lsa_tfs = merge_tfs(ip_dir_names, f'{LSA_RESULTS}{EXT}')
+
+    # If the TFS data contains everything we need: get the imbalance
+    if _validate_for_imbalance(new_data):
```
fsoubelet, line 169: cf validation indent level I mentioned earlier

fsoubelet (changes requested):
See comments, plus 2 points:

- Please add type hints and proper docstrings
- Remove the comment lines and make most of them into proper logging

omc3/scripts/luminosity_imbalance.py
Mael-Le-Garrec: The imbalance is only computed if all 4 files are presented, the script stills merges if that's not the case

JoschD: As Mael said. If it confuses rename to '_all_data_for_imbalance_present()'

JoschD:
Also add to changelog in Readme.md

JoschD:
> Following from Eq(17) in [Concept of Luminosity](https://cds.cern.ch/record/941318/files/p361.pdf), the Luminosity imbalance is given by:
> 
> 
> 
> \frac{L_{IP1}}{L_{IP5}}=\frac{\sqrt{\beta_{x1,IP5}+\beta_{x2,IP5}}\cdot\sqrt{\beta_{y1,IP5}+\beta_{y2,IP5}}}{\sqrt{\beta_{x1,IP1}+\beta_{x2,IP1}}\cdot\sqrt{\beta_{y1,IP1}+\beta_{y2,IP1}}}

Which is what this script is doing. 
Have you calculated the error propagation though?

Mael-Le-Garrec:
> > Following from Eq(17) in [Concept of Luminosity](https://cds.cern.ch/record/941318/files/p361.pdf), the Luminosity imbalance is given by:
> > 
> > \frac{L_{IP1}}{L_{IP5}}=\frac{\sqrt{\beta_{x1,IP5}+\beta_{x2,IP5}}\cdot\sqrt{\beta_{y1,IP5}+\beta_{y2,IP5}}}{\sqrt{\beta_{x1,IP1}+\beta_{x2,IP1}}\cdot\sqrt{\beta_{y1,IP1}+\beta_{y2,IP1}}}
> 
> @lmalina
> Which is what this script is doing.
> Have you calculated the error propagation though?
> 
> @Mael-Le-Garrec I would refrain from linking comments on github PR requests as sources for equations.
> You can write latex in doc-strings that is then converted by sphinx:
> 
> ```
> .. math::
>     \\frac{L_{IP1}}{L_{IP5}}=\\frac{\\sqrt{\\beta_{x1,IP5}+\\beta_{x2,IP5}}\\cdot\\sqrt{\\beta_{y1,IP5}+\\beta_{y2,IP5}}}{\\sqrt{\\beta_{x1,IP1}+\\beta_{x2,IP1}}\\cdot\\sqrt{\\beta_{y1,IP1}+\\beta_{y2,IP1}}}
> ```

why

omc3/scripts/merge_kmod_results.py
```diff
@@ -0,0 +1,244 @@
...
+            f"Computing average betastars and absolute errors for IP {ip}, for both planes "
+            f"and beams"
+        )
+        ip_row = data_frame.loc[data_frame[LABEL].str.startswith(ip)]
+        average_beta = {}
+        error = {}
+
+        for plane in PLANES:
+            average_beta[plane] = 0.5 * ip_row[f"{BETASTAR}{plane}"].sum(axis=0)
+            error[plane] = 0.5 * sqrt((ip_row[f"{ERR}{BETASTAR}{plane}"] ** 2).sum(axis=0))
```
mihofer, line 88: glancing over it, it appears that the formula in the documentation and formulas in the code dont match up .. maybe another sentence that this documentation formulas is correct, implemented here is an approx

lmalina:
> \frac{L_{IP1}}{L_{IP5}}=\frac{\sqrt{\beta_{x1,IP5}+\beta_{x2,IP5}}\cdot\sqrt{\beta_{y1,IP5}+\beta_{y2,IP5}}}{\sqrt{\beta_{x1,IP1}+\beta_{x2,IP1}}\cdot\sqrt{\beta_{y1,IP1}+\beta_{y2,IP1}}}

\sigma_{L_{IP1}/L_{IP5}} = \frac{L_{IP1}}{2 \cdot L_{IP5}} \cdot  \sqrt{
    \sum_{z \in (x,y), i \in (IP1, IP5) }{\frac{\sigma^2_{\beta_{z1,i}} + \sigma^2_{\beta_{z2,i}}} {(\beta_{z1,i}+\beta_{z2,i})^2}} }

JoschD:
> The error of luminosity imbalance is:
> 
> , i \in (IP1, IP5) }{\frac{\sigma^2_{\beta_{z1,i}} %2B \sigma^2_{\beta_{z2,i}}} {(\beta_{z1,i} %2B \beta_{z2,i})^2}} })

If I see this correctly then, our code is not only missing the square-sqrt as I expected, but also a the imbalance itself.

```python
    imbalance = effective_betas[IPS[0]] / effective_betas[IPS[1]]
    relative_error = imbalance * sqrt(sum(relative_errors.values()**2))
```

setup.py
```diff
@@ -30,6 +30,7 @@
     "pytz>=2018.9",
     "h5py>=2.9.0",
     "pytimber>=2.8.0",
+    "uncertainties>=3.1.4",
```
JoschD, line 33: To be discussed xD

tests/unit/test_imbalance.py
```diff
@@ -19,11 +19,11 @@ def test_result_tfs(_tfs_file):
     res = merge_kmod_results.get_lumi_imbalance(_tfs_file)
 
     assert res["imbalance"] == 0.974139299943968
-    assert res["relative_error"] == 0.003859317636164786
-    assert res["eff_beta_ip1"] == 0.39800399369855577
-    assert res["rel_error_ip1"] == 0.0015434779687894644
-    assert res["eff_beta_ip5"] == 0.4085698972636139
-    assert res["rel_error_ip5"] == 0.0023158396673753213
+    assert res["relative_error"] == 0.0027110906825478205
```
JoschD, line 20: did you check if this is compatible with the formula?

tests/unit/test_imbalance.py
```diff
@@ -19,11 +19,11 @@ def test_result_tfs(_tfs_file):
...
+    assert res["eff_beta_ip1"] == 0.3980039936985557
+    assert res["rel_error_ip1"] == 0.0006143103957639417
+    assert res["eff_beta_ip5"] == 0.40856989726361387
+    assert res["rel_error_ip5"] == 0.0009461823749785368
```
JoschD, line 26: Should be renamed to `test_merge_kmod_results.py`  btw

omc3/scripts/merge_kmod_results.py
```diff
@@ -71,39 +71,32 @@ def get_lumi_imbalance(data_frame: tfs.TfsDataFrame) -> Dict[str, float]:
...
-            sum([(error[p] ** 2) / (average_beta[p] ** 2) for p in PLANES])
-        )
-        effective_betas[ip] = sqrt(reduce(mul, average_beta.values()))
+        lumi_coefficient[ip] = 0.5*reduce(mul,
+                                          [up.sqrt(
+                                            up.uarray(ip_row[f"{BETASTAR}{plane}"].values,
+                                                      ip_row[f"{ERR}{BETASTAR}{plane}"].values).sum()
+                                                   )
+                                            for plane in PLANES]
+                                           ) # at some point when omc3 is py>=3.8 this can be replaced by prod()
```
JoschD, line 87: You can use `np.prod` already ... :-\

setup.py
JoschD: (also: are you choosing this version, because it is pi?)

tests/unit/test_imbalance.py
JoschD: I couldn't put comments on the lines, but: 
```
RESULTS = "results.tfs"
LSA_RESULTS = "lsa_results.tfs"
```
should be imported from the script.

-----------

we should have one test that calls the function from commandline to see if the entrypoint-parameters were set properly

JoschD (changes requested):

setup.py
mihofer: > 
> 
> (also: are you choosing this version, because it is pi?)

(nop is the latest, but my fingers are itching to add some comment on the version number)

omc3/scripts/merge_kmod_results.py
```diff
@@ -82,8 +80,8 @@ def get_lumi_imbalance(data_frame: tfs.TfsDataFrame) -> Dict[str, float]:
 
         lumi_coefficient[ip] = 0.5*reduce(mul,
                                           [up.sqrt(
-                                            up.uarray(ip_row[f"{BETASTAR}{plane}"].values,
```
fsoubelet: We should use `.to_numpy()` instead of `.values`

tests/unit/test_imbalance.py
mihofer: there is already one tests with a call from cmdline ..

omc3/scripts/merge_kmod_results.py
```diff
@@ -0,0 +1,237 @@
...
+        A dictionary with the imbalance, the betas at IPs and their errors.
+    """
+
+    lumi_coefficient: Dict[str, float] = {}
+
+    for ip in IPS:
+        LOG.debug(
+            f"Computing lumi contribution from optics for IP {ip}"
+        )
+        ip_row = data_frame.loc[data_frame[LABEL].str.startswith(ip)]
```
JoschD, line 79: it should be `ip_rows` as it is two (one for each beam). Otherwise the `.sum()` below is very confusing.
Maybe even `rows_beams_in_ip`

omc3/scripts/merge_kmod_results.py
lmalina: or better, implement the correct one

setup.py
lmalina: > To be discussed xD

We mostly do our calculations on numpy.arrays or pandas.DataFrame and uncertainties does not seem to be well developed on this front. The performance in terms of CPU time and memory use seems to be orders of magnitude behind that of numpy. The compatibility at least with pandas seems to be questionable. This looks enough not to depend on uncertainties package.

JoschD: I agree with the performance problems and with the pandas problems (which seem to be coming form pandas side). Basic numpy operations work fine though and as we do a lot of `.to_numpy()` anyway this should not be a problem.
I would also not recommend to replace all of our calculations with this uncertainty package, but I see some advantages using it for small scripts like this (where it doesn't matter if the calculation for a 100-number array takes 4us or 400us).

- You can keep errors and values together without mapping them to dictionaries (reduction of variables)
- simple calculations can be performed easily and with errors, without doing the error propagation manually
- significant digits are automatically taken care of by the `__str__()` function

which is basically all this package aims at and it delivers. 
Sure, it would be one more dependency of omc3, but one that I can see be very useful for smaller (i.e. low array length, low repetivity) calculations in the future. 

In case this package is vetoed, I suggest writing a small class that 
- has `.value` and `.error` attributes
- supports implements error propagation for `__add__`, `__sub__`, `__mul__` and `__div__`
- implements `__repr__` as full float and `__str__` as significant digits (like UFloat)
- has a function `round()` which uses the significant digits `float=True` version to round the numbers

tests/unit/test_imbalance.py
JoschD: image

top: full error calculated by uncertainty package
bottom: full error calculated by formular

setup.py
JoschD: Also: 
- errors of in-between-steps are also easily visible e.g. for debug.

Not limited to debug, though: If we want to get rid of the uncertainty package, we will need to calculated the errors for the effective betas also manually...

omc3/utils/iotools.py
```diff
@@ -245,7 +245,7 @@ def __new__(cls, value):
         return Path(value)
 
 
-def convert_paths_in_dict_to_strings(dict_):
+def convert_paths_in_dict_to_strings(dict_: dict) -> dict:
```
fsoubelet, line 248: Beware that this acts on the `dict_` inplace too.

Maybe make a copy first, or simply a temp list / dict that you return in the end.

Then the function will be pure, which we should always strive for.
Type hinting could maybe be improved?

omc3/scripts/merge_kmod_results.py
```diff
@@ -120,7 +131,7 @@ def _check_tfs_sanity(data_frame: tfs.TfsDataFrame):
         raise KeyError(msg)
 
     # check that there is no weird additional data
-    too_many_entries = [ip for ip in IPS if len(data_frame[NAME].str.startswith(ip)) > 2]
+    too_many_entries = [ip for ip in IPS if sum(data_frame.index.str.startswith(ip)) > 2]
```
fsoubelet, line 133: Maybe use `.sum()` from pandas directly for speed?

omc3/utils/iotools.py
```diff
@@ -266,7 +266,7 @@ def convert_paths_in_dict_to_strings(dict_):
     return dict_
 
 
-def remove_none_dict_entries(dict_):
+def remove_none_dict_entries(dict_: dict) -> dict:
```
fsoubelet, line 269: Same here, not pure

omc3/utils/iotools.py
JoschD: Changed it to not return anything and put warning in doc

omc3/scripts/merge_kmod_results.py
JoschD: image

valid point to get it into habit. Otherwise this is like <100 entries here... 

fixed it

tests/unit/test_imbalance.py
fsoubelet: So you mean that uncertainties might not be needed here?

JoschD (approved):

omc3/utils/iotools.py
fsoubelet: I'd rather make the functions pure honestly.

Anything goes wrong in the function and at least you've still got your initial data.
@lmalina thoughts?

omc3/scripts/merge_kmod_results.py
fsoubelet: Still saves about 5 us for 50 elements :p

omc3/scripts/merge_kmod_results.py
```diff
@@ -0,0 +1,275 @@
...
+    Returns:
+         Tuple with the imbalance, the betas at IPs as ufloats.
+    """
+
+    lumi_coefficient: Dict[str, UFloat] = {}
+
+    for ip in IPS:
+        LOG.debug(
+            f"Computing lumi contribution from optics for IP {ip}"
+        )
```
lmalina, line 187: one line?

omc3/scripts/merge_kmod_results.py
```diff
@@ -0,0 +1,275 @@
...
+        data_frame (tfs.TfsDataFrame): a loaded `TfsDataFrame` to validate.
+    """
+    # Check that both beams are there only once
+    multiple_names = [name for name in data_frame.index if (data_frame.index == name).sum() > 1]
+    if multiple_names:
+        msg = (f"Found entries '{', '.join(set(multiple_names))}' "
+               f"several times in merged DataFrame. "
+               "Expected only once")
+        LOG.error(msg)
+        raise KeyError(msg)
```
lmalina, line 130: what about using duplicated method from pandas? something like this?
```
if data_frame.index.is_unique():
    raise KeyError(f"Found duplicated indices '{', '.join(set(data_frame.index.to_list()[data_frame.index.duplicated()]))}")
```

omc3/scripts/merge_kmod_results.py
```diff
@@ -0,0 +1,275 @@
...
+
+    # check that there is no weird additional data
+    too_many_entries = [ip for ip in IPS if data_frame.index.str.startswith(ip).sum() > 2]
+    if too_many_entries:
+        msg = ("More than two entries found for ips "
+               f"{', '.join(too_many_entries)} in merged DataFrame. "
+               "Expected one for each beam.")
+        LOG.error(msg)
+        raise KeyError(msg)
```
lmalina, line 139: is this really needed? Can we have more than two unique indices and care about it at the same time?

omc3/scripts/merge_kmod_results.py
```diff
@@ -0,0 +1,275 @@
...
+def get_lumi_imbalance(data_frame: tfs.TfsDataFrame) -> Tuple[UFloat, UFloat, UFloat]:
+    """
+    Calculate the IP1 / IP5 luminosity imbalance. The luminosity is taken as defined in
+    `Concept of luminosity`, Eq(17): https://cds.cern.ch/record/941318/files/p361.pdf
+
+    The calculation of the luminosity imbalance is then:
+
+    .. math::
+
+        \\frac{L_{IP1}}{L_{IP5}}=\\frac{\\sqrt{\\beta_{x1,IP5}+\\beta_{x2,IP5}}\\cdot\\sqrt{\\beta_{y1,IP5}+\\beta_{y2,IP5}}}{\\sqrt{\\beta_{x1,IP1}+\\beta_{x2,IP1}}\\cdot\\sqrt{\\beta_{y1,IP1}+\\beta_{y2,IP1}}}
```
lmalina, line 172: Also the error?

omc3/scripts/merge_kmod_results.py
```diff
@@ -0,0 +1,275 @@
...
+    new_tfs.headers = headers
+    new_tfs = new_tfs.set_index(NAME)
+
+    # drop BPMWK and check tfs
+    new_tfs = new_tfs.loc[~new_tfs.index.str.startswith(BPMWK), :]
+    _check_tfs_sanity(new_tfs)
+    return new_tfs
+
+
+def get_ip_dir_names(kmod_dirs: List[pathlib.Path]) -> List[pathlib.Path]:
```
lmalina, line 258: There was a comment somewhere, what this method does. Maybe it should have a better name? kmod_results_dirs?

omc3/scripts/merge_kmod_results.py
```diff
@@ -0,0 +1,275 @@
...
+    for d in kmod_dirs:
+        if not d.is_dir():
+            msg = f"Directory {d} does not exist"
+            LOG.error(msg)
+            raise NotADirectoryError(msg)
+
+    pattern = re.compile(".*ip[0-9]B[1-2]")
+    ip_dir_names = [
+        d for kmod in kmod_dirs for d in kmod.glob("**/*") if pattern.match(d.name) and d.is_dir()
+    ]
```
lmalina, line 269: a bit misleading to call `d` always different thing

omc3/scripts/merge_kmod_results.py
```diff
@@ -0,0 +1,275 @@
+"""
+Merge KMOD Results
+-------------------
+
+Script to merge the results from KMOD into one TfsDataFrame.
+The script takes the kmod-results folders as input and merges the
+lsa-result tfs-files in these together.
```
lmalina, line 7: Can be written out in a directory to serve as a measurement for global correction?
Btw, I did not see anywhere anything like "beta_kmod_", maybe I just overlooked it...

omc3/scripts/merge_kmod_results.py
```diff
@@ -0,0 +1,275 @@
...
+Headers of the same name will be overwritten (depending on the alphabetical
+order of the directory names).
+
+Some sanity checks are performed, e.g. that there is only one entry per element.
+If IP1 and IP5 results are given for both planes and beams, the luminosity
+imbalance between these IPs is also calculated and written into the header,
+as well as logged.
+
+The resulting TfsDataFrame is returned, but also written out if an `outputdir`
+is given.
```
lmalina, line 19: important -> to the top

lmalina (changes requested):
Generally, I think we should avoid adding many more things to pull requests. It takes more time to develop, review etc... also we will have less clean history (when every commit changes significant part of repository)

omc3/scripts/merge_kmod_results.py
JoschD: done

omc3/scripts/merge_kmod_results.py
JoschD: done

omc3/scripts/merge_kmod_results.py
JoschD: to be done later in global correct #135

lmalina (approved):

setup.py
fsoubelet: Good points.

I still think this has this place in `PyLHC` rather than `omc3` as we don't need another dependency for the package, especially if only used in small scripts.

omc3/scripts/merge_kmod_results.py
```diff
@@ -0,0 +1,275 @@
...
+*--Optional--*
+
+- **outputdir** *(Path)*:
+
+    Output directory where to write the result tfs
+
+"""
+import pathlib
+import re
+from collections import OrderedDict
```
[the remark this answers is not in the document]
Mael-Le-Garrec: Not completely true, depending on how you use them
 https://stackoverflow.com/questions/50872498/will-ordereddict-become-redundant-in-python-3-7

JoschD (approved):

fsoubelet (approved):

lmalina (approved):

PR #229, opened 2020-09-22
Fixes #227: Changes the direction in model compensation

### LINKED ISSUE #227: Optics measurements: Beam direction in model compensation
Model compensation works for LHC beam 2 as tracked in MADX, while beam 4 should be the required behaviour.   Beam 4 in MADX notation is in reality the beam 2, which is counter rotating with respect to beam1.

lmalina:
counter rotating beams with respect to the model are treated correctly


omc3/optics_measurements/phase.py
```diff
@@ -65,12 +65,14 @@ def calculate(meas_input, input_files, tunes, plane, no_errors=False):
     df = pd.merge(df, input_files.joined_frame(plane, [f"MU{plane}", f"{ERR}MU{plane}"],
                                                dpp_value=dpp_value, how=how),
                   how='inner', left_index=True, right_index=True)
+    df[input_files.get_columns(df, f"MU{plane}")] = (input_files.get_data(df, f"MU{plane}")
```
JoschD, line 68: maybe a good time to clean-up column names?

tests/accuracy/test_measure_optics.py
```diff
@@ -29,7 +29,7 @@ def _create_input(motion):
                     outputdir=BASE_PATH)
     optics_opt, rest = _optics_entrypoint(opt_dict)
     optics_opt.accelerator = manager.get_accelerator(rest)
-    lins = optics_measurement_test_files(opt_dict["model_dir"], dpps, motion)
+    lins = optics_measurement_test_files(opt_dict["model_dir"], dpps, motion, 1)
```
JoschD, line 32: use KW-args. this random `1` does not mean anything.

alternatively: beam_direction could be a string: "clockwise" or "counterclockwise", ("cw" "ccw"? "forward", "backward"?) which is then mapped to 1 and -1 in generate_linfiles ?

JoschD: I just saw that this is only test.. still `beam_direction = 1` here, otherwise ignore my comment

JoschD (changes requested):
You are not testing `beam_direction = -1` ?

omc3/optics_measurements/phase.py
lmalina: not in this PR

JoschD (approved):
constants for column names :'(

PR #237, opened 2020-10-02
Fix nonlinear option problem

mihofer:


tests/accuracy/test_crdt.py
```diff
@@ -45,53 +45,59 @@
...
-    path_to_lin = join(LIN_DIR, order)
-    optics_opt = MEASURE_OPTICS_SETTINGS.copy()
-    optics_opt.update({
-        'files': [join(path_to_lin, f'{order}{idx}') for idx in range(1, 4)],
-        'outputdir': abspath(join(RESULTS_PATH, order)),
-    })
-    hole_in_one_entrypoint(**optics_opt)
-    return (optics_opt, path_to_lin)
 
+@pytest.fixture(scope='module')
```
lmalina, line 52: nice ;-) it does not run run hole_in_one at test collection time anymore

tests/accuracy/test_crdt.py
```diff
@@ -45,53 +45,59 @@
...
+def _create_input():
```
fsoubelet, line 53: Could we not use `pytest`'s own [tmpdir or tmpdir_factory](https://docs.pytest.org/en/stable/tmpdir.html) fixture to make things cleaner?

tests/accuracy/test_crdt.py
```diff
@@ -45,53 +45,59 @@
...
+    omc3_input = {}
+    with tempfile.TemporaryDirectory() as cwd:
+        for order in ORDERS:
+            path_to_lin = join(LIN_DIR, order)
+            optics_opt = MEASURE_OPTICS_SETTINGS.copy()
+            optics_opt.update({
```
fsoubelet, line 59: We should use pathlib everywhere

lmalina (approved):
👍

fsoubelet (changes requested):
I'm gonna be an ass here and insist on using `Pathlib`

mihofer:
> 
> 
> I'm gonna be an ass here and insist on using `Pathlib`

for this PR/crdt tests ok, but switching all other tests to Pathlib should go to a new issue .. also bc e.g. omc3 is not 100% pathlib compatible too (model_dir needs to be string etc.)

lmalina (approved):

fsoubelet (approved):

Issue #238, opened 2020-10-05
switch (completely) to Pathlib


mihofer:
Currently, a mixture of os.path and Pathlib is used across omc3 
old code relying on os.path should be switched and, where possible, functions be made compatible with Paths as input

fsoubelet:
Worth noting: this might require us to take care of this in our other packages first, like I did for `tfs-pandas`.

PR #230, opened 2020-09-23
Feature/GitHub actions

mihofer:
add testing via github actions
Fixes #224

PR #240, opened 2020-10-06
add output dir option to kmod 

### LINKED ISSUE #239: add output dir option to kmod
add an optional outputdir option to the kmod analysis,
if no dir is provided, the inputdir should still be used

test could then be adapted to remove suhtil.rmtree and use pytest tmp_dir instead

mihofer:
Fixes #239

### DIFF
--- a/omc3/run_kmod.py
+++ b/omc3/run_kmod.py
@@ -33,6 +33,9 @@ def kmod_params():
                          type=str,
                          required=True,
                          help='path to working directory with stored KMOD measurement files',)
+    parser.add_parameter(name='output_directory',
+                         type=str,
+                         help='path where outputfiles will be stored, defaults to working_directory',)
     parser.add_parameter(name='beam',
                          type=str,
                          choices=['B1', 'B2'], required=True,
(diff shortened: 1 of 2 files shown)


Mael-Le-Garrec:
For consistency issues you could use `outputdir` instead of `output_directory`

omc3/run_kmod.py
```diff
@@ -33,6 +33,9 @@ def kmod_params():
                          type=str,
                          required=True,
                          help='path to working directory with stored KMOD measurement files',)
+    parser.add_parameter(name='outputdir',
+                         type=str,
```
Mael-Le-Garrec, line 37: why not `type=pathlib.Path` ?

Mael-Le-Garrec (dismissed):

omc3/run_kmod.py
mihofer: switching everything to Pathlib (and testing everything in the background behaves the same) is a separate issue #238 and i think not in the spirit of what this quick fix PR is supposed to do

fsoubelet:
I have implemented this already in the github actions branch to fix builds. How should we proceed?

fsoubelet:
Implemented in #230, will close this once #230 is merged

PR #242, opened 2020-10-09
Readme fix

fsoubelet:
- Re-include the `merge_kmod_results` in the scripts section, that I had forgotten
- Some cleanup, to make the file a bit more lightweight


fsoubelet:
Why do I have so many changes when I only modified the README :(

PR #245, opened 2020-10-09
Fixes #244: Does not pass pandas.Index anymore

### LINKED ISSUE #244: Write only python or numpy types using sdds
On writing of turn_by_turn data we pass `pandas.Index` to `sdds.write()`, which obviously does not know pandas. should be casted to numpy...

lmalina:


fsoubelet (approved):
:+1:

PR #247, opened 2020-10-12
fix option --keep_exact_zeros

### LINKED ISSUE #246: running with --keep_exact_zeros crashes
Running omc3 with the option --keep_exact_zeros crashes with a `ValueError: Index data must be 1-dimensional` 
issue stems from  https://github.com/pylhc/omc3/blob/f530c921deaab5f35a071620ce83bf5e76fef5a7/omc3/harpy/clean.py#L124 as it return a `pd.DataFrame` and not the required empty `pd.Index` 
exact zeros check is also skipped in the accuracy tests

mihofer:
Fixes #246


mihofer:
Currently, the extended test options only include different windows and keep_zeros yes/no .. can be extended (by someone else) if needed/desired

omc3/harpy/clean.py
```diff
@@ -140,14 +140,14 @@ def _get_bad_bpms_summary(harpy_input, known_bad_bpms, bpm_flatness, bpm_spikes,
...
+        raise ValueError("Total Number of BPMs after filtering is zero.")
     n_bad_bpms = n_total_bpms - n_good_bpms
     LOGGER.debug(f"(Statistics for file reading) Total BPMs: {n_total_bpms}, "
                  f"Good BPMs: {n_good_bpms} ({(100 * n_good_bpms / n_total_bpms):2.2f}%), "
                  f"Bad BPMs: {n_bad_bpms} ({(100 * n_bad_bpms / n_total_bpms):2.2f}%)")
     if (n_good_bpms / n_total_bpms) < 0.5:
         raise ValueError("More than half of BPMs are bad. "
-                         "This could be cause a bunch not present in the machine has been "
+                         "This could be because a bunch not present in the machine has been "
                          "selected or because a problem with the phasing of the BPMs.")
```
lmalina, line 151: then also "because of a problem"

omc3/harpy/clean.py
```diff
@@ -32,7 +32,7 @@ def clean(harpy_input, bpm_data, model):
     """
     bpm_data, bpms_not_in_model = _get_only_model_bpms(bpm_data, model)
     if bpm_data.empty:
-        raise AssertionError("Check BPMs names! None of the BPMs was found in the model!")
+        raise AssertionError("Check BPMs names! None of the BPMs were found in the model!")
```
lmalina, line 35: This is wrong, change it back.

lmalina (changes requested):

omc3/harpy/clean.py
mihofer: https://data.grammarbook.com/blog/singular-vs-plural/none-were-vs-none-was/

lmalina: This exists only in US english.

lmalina (approved):
Putting English aside.

PR #248, opened 2020-10-12
Use dateutil instead of pytz

### LINKED ISSUE #226: time_tools: replace pytz with dateutil.tz
See https://blog.ganssle.io/articles/2018/03/pytz-fastest-footgun.html

Should hopefully be an easy rewrite without hiccups.
Be sure to test the `fold` functionality.

Mael-Le-Garrec:
Fixes #226 

dateutil.tz works a bit differently than pytz, the `localize` function for instance can't be used anymore.
All datetimes are here zone-aware and one change to be noted is that a AccDatetime object can now be created directly with a `tzinfo` instead of using the `.local` property on it.
`dt.astimezone` returns a new object, so does `dt.replace`, we would always get a AccDatetime object in UTC.

### DIFF
--- a/omc3/utils/time_tools.py
+++ b/omc3/utils/time_tools.py
@@ -12,6 +12,7 @@ in particular to switch easily between local time and utc.
 from datetime import datetime, timedelta
 
 import pytz
+import dateutil.tz as tz
 
 from omc3.definitions.formats import TIME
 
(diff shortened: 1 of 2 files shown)


JoschD (dismissed):
Looks good to me. I'm not sure if it's not now in some place a little more complicated than it need to be (as some of the functions arose from the way timezones worked before), but as long as the tests are passing ...

JoschD (approved):
Sorry, I only wanted to check the fold-test, as I was confused by it. As Above: Approved

JoschD (dismissed):
Aller guten Dinge sind drei

omc3/utils/time_tools.py
```diff
@@ -101,17 +107,20 @@ def __new__(cls, *args, **kwargs):
             dt = args[0]
         else:
             dt = datetime.__new__(cls, *args, **kwargs)
-        dt = utc_to_local(dt, pytz.utc)  # does not convert, but checks tz
+
+        if dt.tzinfo == None:
```
fsoubelet, line 111: would be better to use `is None` here, as `== None` can sometimes return a non-boolean

fsoubelet (approved):

PR #251, opened 2020-10-14
Fix set-env use

fsoubelet:
Should fix the `set-env` use in the `coverage` workflow.
Hopefully fix the warning in the `documentation` workflow by updating the used version of the `github-pages-deploy-action`.

Need to open this PR to trigger the first one.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -53,16 +53,16 @@ jobs:
 
       - name: Set up env for CodeClimate (push)
         run: |
-          echo "::set-env name=GIT_BRANCH::$GITHUB_REF"
-          echo "::set-env name=GIT_COMMIT_SHA::$GITHUB_SHA"
+          echo "GIT_BRANCH=$GITHUB_REF" >> $GITHUB_ENV
+          echo "GIT_COMMIT_SHA=$GITHUB_SHA" >> $GITHUB_ENV
         if: github.event_name == 'push'
 
       - name: Set up env for CodeClimate (pull_request)
         env:
           PR_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
         run: |
-          echo "::set-env name=GIT_BRANCH::$GITHUB_HEAD_REF"
-          echo "::set-env name=GIT_COMMIT_SHA::$PR_HEAD_SHA"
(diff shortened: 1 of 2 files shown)


fsoubelet:
That fixed the issue for the `coverage` workflow: as can be seen [here](https://github.com/pylhc/omc3/actions/runs/306506145), no more warnings.

I think we can always merge this PR and see what happens with the `documentation` workflow. Eventually, move to a different gh-pages deploy action or open an issue to the maintainers or the ones we currently use, as I'm fairly sure it is the source of our warning.

@lmalina @mihofer thoughts?

mihofer:
Fine with me, looking at the deploy [repo](https://github.com/JamesIves/github-pages-deploy-action/issues) i couldn't find anything related to that issue since the last version though ...
referencing a specific version by number is necessary or could one just always refer to the latest version to avoid PRs whenever a package is updated? (but likely not needed very often)

mihofer (approved):

fsoubelet:
> looking at the deploy repo i couldn't find anything related to that issue since the last version though

Well indeed, but I really hope the issue comes from them considering I can't find any other possible use of `set-env` in our `documentation` workflow.

> referencing a specific version by number is necessary or could one just always refer to the latest version to avoid PRs whenever a package is updated? (but likely not needed very often)

Looking at the relevant [documentation](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-using-versioned-actions) shows we have many options, and we could for instance choose to specify `github-pages-deploy-action@v3` which would use the latest `3.x` version available. We can also specify a branch, a SHA, etc.

However, adopting such behavior might be to be discussed with the team first.

mihofer:
> 
> Looking at the relevant [documentation](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-using-versioned-actions) shows we have many options, and we could for instance choose to specify `github-pages-deploy-action@v3` which would use the latest `3.x` version available. We can also specify a branch, a SHA, etc.
> 
> However, adopting such behavior might be to be discussed with the team first.

well, actually reads there any way that version should be included 

> We strongly recommend that you include the version of the action you are using by specifying a Git ref, SHA, or Docker tag number. If you don't specify a version, it could break your workflows or cause unexpected behavior when the action owner publishes an update.

so better leave it as is

Issue #272, opened 2021-02-25
Setup extra dependencies for technical network fonctionality


fsoubelet:
Since January, `pyjapc`, `cmmnbuild-dep-manager`, `pjlsa`, `jpype1` and `pytimber` are not installable from `PyPI` anymore, and should be installed directly from the `acc-py` PI. The corresponding packages found on `PyPI` are placeholders left there to avoid shadow clones and are setup to fail at the installation step (see error below for `pytimber` for instance).

```
ERROR: Command errored out with exit status 1:
     command: /opt/hostedtoolcache/Python/3.8.7/x64/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-l60xkecp/pytimber_4ea8fc7596ab4645b2ea167b2ff7d93d/setup.py'"'"'; __file__='"'"'/tmp/pip-install-l60xkecp/pytimber_4ea8fc7596ab4645b2ea167b2ff7d93d/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-z74pg7pe
         cwd: /tmp/pip-install-l60xkecp/pytimber_4ea8fc7596ab4645b2ea167b2ff7d93d/
    Complete output (13 lines):
    
    ***
    
    Please install the `pytimber` package
    
    using the https://acc-py-repo.cern.ch package index.
    
    Additional information can be found in:
    
    https://wikis.cern.ch/pages/viewpage.action?pageId=145493385
    
    ***
```

The proper command with`pip` would then be:
```bash
pip install --index-url https://acc-py-repo.cern.ch/repository/vr-py-releases/simple --trusted-host acc-py-repo.cern.ch pytimber
```

However, since the `acc-py` PI is not accessible from outside the CERN network, that is not a solution for our CI.
As discussed today with Phil Elson, the best solution for now would be to declare these as optional dependencies, which will be fine for installation, and fine when deploying to `PyPI` as long as the placeholders are found there.

Checklist:
- [ ] Check that CI workflows work fine.
- [ ] Document the caveat.
[...]

PR #273, opened 2021-02-25
Tech dependency for pytimber

fsoubelet:
This is the implementation PR for https://github.com/pylhc/omc3/issues/272

Checklist:
- [x] Check that CI workflows install fine.
- [x] Document the caveat.
- [x] Failsafe the imports of `pytimber` functionality in `omc3` (currently only in `omc3.tune_analysis.timber_extract.py`, which is only imported  in `omc3.amplitude_detuning_analysis.py`).
- [x] Specify `source="nxcals"` when declaring `db = pytimber.LoggingDB()`, since the change from `CALS` to `NXCALS` has been made.

I took the opportunity to remove unsused parts of `setup.py` and do a little bit of type hinting in `omc3.tune_analysis.timber_extract.py`.

I also named the extra `tech` to stay consistent with the naming adopted in `PyLHC`.

### DIFF
--- a/omc3/tune_analysis/timber_extract.py
+++ b/omc3/tune_analysis/timber_extract.py
@@ -2,15 +2,13 @@
 Timber Extraction
 -----------------
 
-Tools to extract data from ``Timber``.
+Tools to extract data from ``Timber``. It is a bit heavy on the LHC side at the moment.
 
-Please note: this module requires the ``pytimber`` package to access ``Timber`` functionality,
+**Please note**: this module requires the ``pytimber`` package to access ``Timber`` functionality,
 both of which are only possible from inside the CERN network.
 
-To install ``pytimber`` along ``omc3``, please do so from inside the CERN network by using the `tech` extra
-dependency (`pip install omc3[tech]`).
-
-It is a bit heavy on the LHC side at the moment.
+To install ``pytimber`` along ``omc3``, please do so from inside the CERN network by using the [tech] extra
+dependency (``pip install omc3[tech]``).
 """
 import re
 
--- a/omc3/tune_analysis/timber_extract.py
+++ b/omc3/tune_analysis/timber_extract.py
@@ -4,6 +4,12 @@ Timber Extraction
 
 Tools to extract data from ``Timber``.
 
+Please note: this module requires the ``pytimber`` package to access ``Timber`` functionality,
+both of which are only possible from inside the CERN network.
+
+To install ``pytimber`` along ``omc3``, please do so from inside the CERN network by using the `tech` extra
+dependency (`pip install omc3[tech]`).
+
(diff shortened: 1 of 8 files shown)


fsoubelet:
In the end there is now a (tested & documented) `omc3.utils.mock` module providing a convenience function to import the problematic packages such as `pytimber` and a mock class returned if they are absent.

Usage is simple as:
```python
from omc3.utils.mock import cern_network_import
pytimber = cern_network_import("pytimber")
db = pytimber.LoggingDB(source="nxcals")  # will raise if pytimber not installed
```

fsoubelet:
To be discussed: maybe renaming the `tech` extra to `cern`. Feasible in `PyLHC` too since it has not been deployed yet to PyPI with this extra (`omc3` is not being deployed is blocking).

Opinions @JoschD @lmalina @mihofer ?

JoschD:
I think I'd go for `cern` as the name for the imports, instead of `tech`. Otherwise really nice!

fsoubelet:
Also, I forgot: credits to @JoschD for the idea of the mock class, and 99% of the implementation ;)

omc3/tune_analysis/timber_extract.py
```diff
@@ -2,29 +2,35 @@
 Timber Extraction
 -----------------
 
-Tools to extract data from ``Timber``.
+Tools to extract data from ``Timber``. It is a bit heavy on the LHC side at the moment.
```
Mael-Le-Garrec, line 5: Did you put an issue or something somewhere to remember to remove this comment when things will go back to a normal state?

fsoubelet: The comment is not from me and has only been moved a few lines up, so I'm not sure. As far as I know this code is initially from @JoschD

JoschD: I only have LHC experience with Timber and I am not sure if we will use it with any other machine soon. It's only for BBQ-cleaning anyway. So LHC is the "normal state"

JoschD: It is meant as a comment: If you have any other machine and you think you can extract data from timber for it... you need to implement that first.

fsoubelet:
As we seem to agree (cf Mattermost), I'm changing in 74cf093 the extra from `tech` to `cern` for the reasons mentioned above. I will make the same change in `PyLHC` when reworking the extra properly there, for consistency.

lmalina (commented):
Nice work, looks good to me...I did not try the actual installation processes though...so just commenting

JoschD:
> Nice work, looks good to me...I did not try the actual installation processes though...so just commenting

I did. Everything seems to install :)

JoschD (approved):

PR #274, opened 2021-02-28
Corrections and Response Matrix

JoschD:

PR #278, opened 2021-03-23
Fix cache step in workflows

fsoubelet:
These changes include `pip`'s out-of-directory cache in the github actions caching step, which will speed up the `Install package` step.

Previous implementation was my bad.


fsoubelet:
As can be seen in the checks below, the basis tests for instance are now successful in around 2 minutes compared to the 6-8 minutes currently in https://github.com/pylhc/omc3/pull/274

JoschD (approved):
I have no idea what is going on, but it looks good

PR #286, opened 2021-06-23
Feature/model creator

### LINKED ISSUE #283: Model Creator: Modifers not created
A `modifiers.madx` file is assumed, when loading model from directory, 
but this file is never written out by the creator. 
Instead the path to the modifiers is inserted into the `job.madx`.

Also there is no problem allowing multiple modifier files. 

possible solutions: 
 - read modifier files from `job.madx` by a common tag
 - create `modifiers.madx` file and read modifiers from there.

### LINKED ISSUE #284: Model Creator: Allow only nat-tunes given if no excitation is selected
This will drop the 
```
if (acd == 0) ...
```
parts of the `job.madx`, which makes it less flexible.

### LINKED ISSUE #285: Model Creator: Create and use accel.get_base_madx_script
The `get_base_madx_script` function is mandatory for all accelerators and will be used by the fullresponse creation.
So far it is only implemented for `lhc`.

JoschD:
Fixes #283 
Fixes #284 
Fixes #285 
Changes towards #238 for model creation, accelerator class and where they were used (e.g. kmod-analysis, madx_wrapper)

Hint: In kmod I also changed the `--beam` input from `"B1"` or `"B2"` to `1` or `2`, to be aligned with `--beam` in accelerator class


omc3/model/accelerators/accelerator.py
```diff
@@ -285,3 +320,23 @@ class AcceleratorDefinitionError(Exception):
...
+
+
+# Helper ----
+
+def _get_modifiers_from_modeldir(model_dir: Path):
+    """Parse modifiers from job.create_model.madx or use modifiers.madx file."""
+    job_file = model_dir / JOB_MODEL_MADX
+    if job_file.exists():
+        job_madx = job_file.read_text()
+        modifiers = re.findall(fr"\s+call,\s*file\s*=\s*[\"\']?([^;\'\"]+)[\"\']?\s*;\s*{MODIFIER_TAG}", job_madx,
```
fsoubelet, line 332: Could use a comment to explain that regex, so we don't need to think about it too much later on

omc3/model/accelerators/lhc.py
```diff
@@ -240,19 +240,19 @@ def get_ips(self):
                    Lhc.DOROS_IP_BPMS.format(side="R", ip=ip, beam=self.beam))
 
     def log_status(self):
-        LOGGER.info(f"  model dir = {self.model_dir}")
-        LOGGER.info("Natural Tune X      [{:10.3f}]".format(self.nat_tunes[0]))
-        LOGGER.info("Natural Tune Y      [{:10.3f}]".format(self.nat_tunes[1]))
-        LOGGER.info("Best Knowledge Model     [{:>10s}]".format(
+        LOG.info(f"  model dir = {self.model_dir}")
+        LOG.info("Natural Tune X      [{:10.3f}]".format(self.nat_tunes[0]))
```
fsoubelet, line 244: let's use f-strings everywhere here

omc3/model/accelerators/ps/base.mask
```diff
@@ -79,42 +79,3 @@ match, use_macro;
...
 ENDMATCH;
-
-
-
-/******************************************************************************************
- * TWISS
- ******************************************************************************************/
-
-
-/************************************************************
```
fsoubelet, line 90: Why are we getting rid of this block?

fsoubelet: Nevermind I see it's moved in a different file.

omc3/model_creator.py
```diff
@@ -20,8 +20,10 @@
 from omc3.model.model_creators.psbooster_model_creator import PsboosterModelCreator
 from omc3.model.model_creators.segment_creator import SegmentCreator
 from omc3.utils.iotools import create_dirs
+from omc3. utils import logging_tools
```
fsoubelet, line 23: typo space in import

omc3/model_creator.py
```diff
@@ -37,15 +39,23 @@
...
-                         help="Output path for model, twiss files will be writen here.")
-    params.add_parameter(name="writeto", type=str,
-                         help="Path to the file where to write the resulting MAD-X script.")
-    params.add_parameter(name="logfile", type=str,
-                         help=("Path to the file where to write the MAD-X script output."
-                               "If not provided it will be written to sys.stdout."))
+    params.add_parameter(
+        name="type",
+        choices=("nominal", "best_knowledge", "coupling_correction"),
+        help="Type of model to create, either nominal or best_knowledge"
```
fsoubelet, line 45: Help does not reflect choices

omc3/model_creator.py
```diff
@@ -110,26 +115,18 @@ def create_instance_and_model(opt, accel_opt):
 
         JPARC: Not implemented
     """
-    if sys.flags.debug:
```
fsoubelet, line 113: Removing the option to enable debug logging? :(

fsoubelet: Nevermind again, I see the logic has been moved to logging_tools 👍

tests/unit/test_model_creator.py
```diff
@@ -1,86 +1,155 @@
...
-                    writeto=join(BASE_OUTPUT, "job.twiss_best_knowledge.madx"),
-                    logfile=join(BASE_OUTPUT, "madx_log_best_knowledge.txt"))
-    create_instance_and_model(opt_dict)
-    _clean_up(BASE_OUTPUT)
-
-
-def _clean_up(path_dir):
-    if isdir(path_dir):
-        rmtree(path_dir, ignore_errors=True)
+def test_lhc_creation_best_knowledge(tmp_path):
```
fsoubelet, line 103: This test does not test the creation.

fsoubelet (changes requested):
Just a small omission as I can tell, otherwise LGTM

tests/unit/test_model_creator.py
JoschD: Ooops. It did at some point

tpersson (approved):
I had a look and no comments on the changes but just observations that still are a lot of things missing such as a real creation of the segment-by-segment but that is fine and I will have a look at that shortly. I think it is crucial that this PR is merged so we can merge the correction branch.

fsoubelet (approved):

PR #287, opened 2021-06-24
Feature/madx logging example to convince felix

JoschD:
* added MADX-level logging
* convinced Felix that this is the way
* pathlib in madx-wrapper tests


fsoubelet (approved):
Second point was clearly the bulk of the work 🙏 

To be clear: the `madx_wrapper` will not log the `MAD-X` output to the command line as `INFO` level anymore after this PR.

If the user activates the Python `-d` debug flag, then `MAD-X` output is logged with level "MADX".

PR #289, opened 2021-07-05
Enhancement/30/coupling

awegsche:
## Coupling in omc3 - Part 1

this is the first partly coupling pull request ... there is more to come in the future

This pull request brings bare bones coupling measurements and rescaling compensation


tpersson:
The coupling part looks nice but there seems to be a problem with the dictionary and in particular, it can't find the key: Free.

tpersson (approved):
Very nice!

tpersson:
If you just merge it and make sure it still works then I think we should go ahead and merge it into the master.

tpersson (approved):

tpersson (approved):

Issue #302, opened 2021-07-22
BUG: model creator hangs in MAD-X process


fsoubelet:
As of last commit, the `model_creator` hangs in the `MAD-X` process as `MAD-X` itself does not find the input file.
The error message logged is 
```
MADX | ++++++ Error: invalid input filename  :  lhc_model/job.create_model.madx | .madx_wrapper
```
As simple reproduction would be the example command from the OMC website (keep the `-d` flag):
```bash
python -d -m omc3.tbt_converter \
    --files trackone \
    --tbt_datatype trackone \
    --drop_elements LHCB1MSIA.EXIT.B1_P_ \
    --outputdir .
```

PR #303, opened 2021-07-22
Model Creator Fix

fsoubelet:
This is a fix for #302 

In `madx_wrapper`, using absolute paths to call the jobfile and `MAD-X` executable ensures the file is found regardless of having `cwd`-ed into the `outputdir`.


JoschD (dismissed):
A few too many changes. Could be two PRs, but I will accept it for now.

JoschD (approved):

PR #308, opened 2021-07-30
Sbs phase advance

tpersson:
This only contains the phase advance correction for the segment-by-segment.

There is a conflict with the master most likely due to the new way of dealing with relative path. Maybe @JoschD or @fsoubelet can guide me on how to change this or just push the needed changes.

(I had some issues with adding files to the other PR so I simply made a new one. )


JoschD:
Will be implemented in #474

PR #311, opened 2021-08-04
Change to relative paths

### LINKED ISSUE #310: Response Creation MAD-X with relative model dir
Response creation does not work if the model-dir is given as a relative path.

JoschD:
Paths now relative in madx-script as we switch to model-dir in madx-run anyway.
Fixes #310.


JoschD:
On a side-note: the response creator function could also be re-written as a ModelCreator one day.

fsoubelet (approved):

Issue #292, opened 2021-07-12
model creation


awegsche:
### add `Fetcher` classes / functions that link lattice definition (madx, sequence, strength files ...) to the model directory:
- `PathFetcher`: user specifiies path, this path is then linked to the model directory
- `AfsFetcher`: user specifies opitcs year (2018, 2022, etc) which is then linked to the model directory
- `LsaFetcher`: looks up info from LSA
### help commands to inform the user about possibilities and workflows
- list command args / usage
```
python omc3/model_creator.py \
    --accel lhc --type nominal --beam1 \
    --fetch afs --show_help
```
possible output:
```
AfsFetcher: usage
Links optics from afs. Available options:
run / year / tune definition from accelerator class
--optics_file        choose the optics
--list opticsfiles   shows a list of available opticsfiles for given settings
                     (run and year have to be set)
--show_help          show this help message and exit without creating a model
*** dry-run: no model created ***
```

- list possible modifers

```
python omc3/model_creator.py \
    --accel lhc --type nominal --beam1 \
    --fetch afs --list_modifiers
```
possible output:
```
AfsFetcher: list modifiers
opticsfile.1
opticsfile.30
opticsfile.30_ctpps1
*** dry-run: no model created ***
```

JoschD:
Related to model creation:
- when to use `match_tunes` and `match_tunes_ats` macros (seems the former does not work for 2018+ optics, even non ATS?)
- when to use `coupling_knobs` and `coupling_knobs_ats`?  Are the definitions there still valid?

mihofer:
> Related to model creation: How to handle the macros.
> Examples :
> 
> * when to use `match_tunes` and `match_tunes_ats` macros (seems the former does not work for 2018+ optics, even non ATS?)
> * when to use `coupling_knobs` and `coupling_knobs_ats`?  Are the definitions there still valid?

this is mostly/only relevant for LHC 
sometime between 2017 and 2018 (and earlier for some other optics) variable names were changed and broke some macros 
one could extend the macros with more if year/optics/md but that might become quite messy quite soon so rather we could move the lhc macros to the directories where the `seq` are stored and keep then keep macros names and interface the same (hopefully)

awegsche:
`lhc_model_creator` looks for modifiers in the model dir:

https://github.com/pylhc/omc3/blob/3272f6f11e11bfe576a4c0615a69734c0064c591/omc3/model/model_creators/lhc_model_creator.py#L112-L117

[...]

PR #316, opened 2021-08-15
Added 2021 strengths

JoschD:
This adds the strengths files for the 2021 model of the PS and changes model creation a bit (YEAR was hardcoded).
This is a temporary fix until #292 is implemented


JoschD:
This  PR looks worse than it is. Only ps.py is changed. The others are the new strengths files

JoschD:
The same should be done for PSBooster (i.e. adding a year)

mihofer (approved):

PR #321, opened 2021-09-11
Adding the sequence for 2021 and 2022

tpersson:
This PR adds the sequence for 2021 and 2022. It follows the same principle as before and it also updates the bpm list using the new sequence. 

If we could merge this as soon as possible that would be great!


fsoubelet:
What is the difference between the `beam[12].tfs` and `beam[12]bpms.tfs`? They are not exactly the same and I don't think we need both.

Do the 2021 files correspond to beam test?

tpersson:
@fsoubelet You are right I, one the beam[12].tfs was added by mistake. I just copied it to the directory but forgot to change it to the correct name.

fsoubelet (approved):
Did not check the files but I assume they're from the AFS optics :)

PR #323, opened 2021-09-15
Modelcreation for beamtest

awegsche:
adds a symlink to acc-models

### DIFF
--- a/omc3/model/accelerators/lhc.py
+++ b/omc3/model/accelerators/lhc.py
@@ -132,7 +132,7 @@ class Lhc(Accelerator):
             name="year",
             type=str,
             required=True,
-            choices=("2012", "2015", "2016", "2017", "2018","2021","2022", "hllhc1.3"),
+            choices=("2012", "2015", "2016", "2017", "2018", "2022", "hllhc1.3"),
             help="Year of the optics (or hllhc1.3).",
         )
         params.add_parameter(name="ats", action="store_true", help="Marks ATS optics")
--- a/omc3/model/model_creators/lhc_model_creator.py
+++ b/omc3/model/model_creators/lhc_model_creator.py
@@ -6,6 +6,7 @@ This module provides convenience functions for model creation of the ``LHC``.
 """
 import logging
 import shutil
+import os
 from pathlib import Path
 from typing import List
 
@@ -30,6 +31,8 @@ from omc3.model.constants import (
 )
 from omc3.model.model_creators.abstract_model_creator import ModelCreator
 from omc3.utils import iotools
+
+ACCMDLS = "/afs/cern.ch/eng/acc-models/lhc"
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


tpersson (approved):
Very nice! Needed changes to be compatible with the new repository structure.

omc3/model/accelerators/lhc.py
```diff
@@ -132,7 +132,7 @@ def get_parameters():
             name="year",
             type=str,
             required=True,
-            choices=("2012", "2015", "2016", "2017", "2018","2021","2022", "hllhc1.3"),
+            choices=("2012", "2015", "2016", "2017", "2018", "2022", "hllhc1.3"),
```
mihofer, line 135: why was 2021 added just two days before to be removed now again?

omc3/model/model_creators/lhc_model_creator.py
```diff
@@ -82,6 +85,13 @@ def get_correction_check_script(
 
     @classmethod
     def prepare_run(cls, accel: Lhc) -> None:
+        print(f"year: {accel.year}")
```
mihofer, line 88: replace with LOGGER.info or LOGGER debug

omc3/model/model_creators/lhc_model_creator.py
```diff
@@ -82,6 +85,13 @@ def get_correction_check_script(
...
+        if accel.year in ["2018", "2022"]:  # this should be handled by the fetcher
+            symlink_dst = Path(f"{accel.model_dir}/acc-models-lhc")
+            print(f"symlink dst: {symlink_dst}")
+            if symlink_dst.is_symlink():
+                os.unlink(symlink_dst)
+            os.symlink(f"{ACCMDLS}/{accel.year}", symlink_dst.absolute())
```
mihofer, line 94: why os when pathlib can do the same?

omc3/model/model_creators/lhc_model_creator.py
```diff
@@ -82,6 +85,13 @@ def get_correction_check_script(
 
     @classmethod
     def prepare_run(cls, accel: Lhc) -> None:
+        print(f"year: {accel.year}")
+        if accel.year in ["2018", "2022"]:  # this should be handled by the fetcher
+            symlink_dst = Path(f"{accel.model_dir}/acc-models-lhc")
+            print(f"symlink dst: {symlink_dst}")
```
mihofer, line 91: also LOGGER

omc3/model/model_creators/lhc_model_creator.py
awegsche: should probably be removed

omc3/model/model_creators/lhc_model_creator.py
awegsche: why pathlib when os can do the same?

mihofer: because it was used in this script before, so in order to keep consistent, avoid extra imports, and any potential future incompatibilities between those modules .. who is to say that the unlink still works in 3 years with a path provided

omc3/model/accelerators/lhc.py
tpersson: Because I added it since I thought it was the way the optics repository was setup but in the end it wasn't so removing it is the right thing.

PR #326, opened 2021-09-17
SbS like jaime

JoschD:
This is the Segment-by-Segment algorithm as envisioned by @jaimecp89.
In this state it should contain the same functionality as the SbsPhaseAdvance (#308) branch, but lacks a test, some doc-strings and some cleanup.

**Advantages**:
 - Separated model creation, segment input and segment-by-segment propagation
 - Object Oriented:
     - One only needs to add a `Propagable` object to add a parameter, which can then be just added to the loop.
     - Highly shared functionality between parameters
 - Uses the Optics-Measurements TfsCollection
 - Allows for Element-Segments as input (see next point)
 - Finds BPMs around from model

**Caveats**:
- Standardized and simplified output names
- Standardized `ini_r##` and `end_r##` to `r##_ini` and `r##_end` (as all the other parameters)
- Renamed `front` (or omission of it) to `forward` and `back` to `backward` everywhere
- Some changes in the Model Creation 
   - some paths (model-dir) 
   - call-logic: see `full_run` function and `post_run` function for checks

**Todo**:
- [x] Move `segment_by_segment.py` to main folder.
- [x] move the `segment_by_segment/matcher` and `sbs_matcher.py`  to new branch (accounts for a lot of changed files, make new issue to adapt it to python3)
- [x] remove debug logging
- [ ] Add Test (see text from Tobias below)
- [ ] Add doc-strings and type-hints

---------------------------------------------
From Tobias :
> I put an example to start testing the sbs on my work public /afs/cern.ch/work/t/tpersson/public/forSBStest_ACdipole/
>
> There are some files also to generate other tracking but the important just to test the SBS are the files in the measured optics such as total_phase_x.tfs etc. You can compare the result to the output in sbs folder there. The  starting point used were:
> 
> BPM.12L1.B1
> BPM.12R1.B1
>
>  And the correction used for the sbs is in the file: sbs/corrections_IP1.madx
>
 > The modifier used was the opticsfile.22.
>
> This is a good first test but then the error propagation also needs to be checked. For my version I checked this by analyzing files from Beta-Beat.src and then compared it to the OMC3 with the option beta-beat.src “names”.


omc3/sbs_matcher.py
```diff
@@ -0,0 +1,233 @@
...
+
+def _run_gui_anaconda():
+    from subprocess import call
+    if sys.platform != "darwin":  # This is Mac
+        if "win" in sys.platform:
+            LOGGER.error("There is not Windows version of Anaconda in OMC.\
+                         Aborting.")
+            return
+    interpreter = os.path.join("/afs", "cern.ch", "eng", "sl", "lintrack",
+                               "omc3_python", "bin", "python")
```
mihofer, line 193: move to constant and put together in a file with acc-models repo/other external dependencies

omc3/segment_by_segment/matcher/gui/matchers_models/matcher_model_amp.py
```diff
@@ -0,0 +1,112 @@
...
+            outpath, "sbs",
+            "sbsampbetabeat" + plane.lower() + "_" + label + ".out")
+        )
+        return file_data
+
+    def _get_beta_beat_file(self, label, plane):
+        outpath = self._matcher_model.get_output_path()
+        file_data = tfs_pandas.read_tfs(os.path.join(
+            outpath, "sbs",
+            "sbsbetabeat" + plane.lower() + "_" + label + ".out")
```
mihofer, line 71: grrr

omc3/segment_by_segment/matcher/gui/matchers_models/matcher_model_default.py
```diff
@@ -0,0 +1,237 @@
...
+import shutil
+import logging
+
+LOGGER = logging.getLogger(__name__)
+
+
+class MatcherModelDefault(object):
+
+    BETA_CORR_CLASSES = ("MQX", "MQXT", "MQXNEW", "MQT", "MQM", "MQTL")
+    COUP_CORR_CLASSES = ("MQSX")
```
mihofer, line 12: very lhc specific, could be rethought at some point

omc3/optics_measurements/io_filehandler.py
```diff
@@ -0,0 +1,45 @@
...
+    KMOD_BETA_NAME, KMOD_BETASTAR_NAME, F1001_NAME, F1010_NAME, EXT
+)
+from tfs.collection import TfsCollection, Tfs
+
+
+class OpticsMeasurement(TfsCollection):
+    """Class to hold and load the measurements from GetLLM.
+
+    The class will try to load the _free file, then the _free2 and then the
+    normal file, if none of them if present an IOError will be raised.
```
JoschD, line 13: Not true anymore

JoschD:
Branch diverged too much from master. Created new branch and PR #474

PR #332, opened 2021-09-29
To enable model creation for 2022 if no afs is available.

tpersson:
Before it was trying to unlink if it was there but now it skips the linking if the folder is already there. This enables the creation of model if you have manually copied or checked out the repository.

### DIFF
--- a/omc3/model/model_creators/lhc_model_creator.py
+++ b/omc3/model/model_creators/lhc_model_creator.py
@@ -87,7 +87,7 @@ class LhcModelCreator(ModelCreator):
     def prepare_run(cls, accel: Lhc) -> None:
         if accel.year in ["2018", "2022"]:  # these years should be handled by the fetcher
             symlink_dst = Path(accel.model_dir)/LHC_REPOSITORY_NAME
-            if symlink_dst.exists() is False:
+            if not symlink_dst.exists():
                 LOGGER.debug(f"Symlink destination: {symlink_dst}")
                 symlink_dst.absolute().symlink_to((ACCELERATOR_MODEL_REPOSITORY/f"{accel.year}"))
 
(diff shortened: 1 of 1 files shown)


mihofer (approved):

fsoubelet (dismissed):
I approve too. In the future I would like this fetcher to be optional, or at least not to kick in if a modifier file has been manually given, so as not to block model creation for machines without AFS access.

omc3/model/model_creators/lhc_model_creator.py
```diff
@@ -87,10 +87,10 @@ def get_correction_check_script(
     def prepare_run(cls, accel: Lhc) -> None:
         if accel.year in ["2018", "2022"]:  # these years should be handled by the fetcher
             symlink_dst = Path(accel.model_dir)/LHC_REPOSITORY_NAME
-            LOGGER.debug(f"Symlink destination: {symlink_dst}")
-            if symlink_dst.is_symlink():
-                symlink_dst.unlink()
-            symlink_dst.absolute().symlink_to((ACCELERATOR_MODEL_REPOSITORY/f"{accel.year}"))
+            if symlink_dst.exists() is False:
```
awegsche, line 90: just as comment, not that I care much: this should be `if not symlink_dst.exists()`

tpersson: This checks for a symlink and also for a normal directory in case you have actually copied the directory to that location.

awegsche (approved):

omc3/model/model_creators/lhc_model_creator.py
JoschD: what andreas means is this: 
image
(https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations)

awegsche: yes, that's what I meant.
in the case "test if false" you don't really lose with this, since a `not` operation is needed anyway. but it looks odd

tpersson: I see! I fixed it in the code now but seems that I need two people to approve again!

mihofer (approved):

fsoubelet (approved):

PR #335, opened 2021-10-01
Fix Amplitude error for RDT when using one file

mihofer:
Before, when using a single file, the error on the RDT amplitude would be inf.
Now, if the fitting returns an infinite errors, it is replaced with 0..

### DIFF
--- a/omc3/optics_measurements/rdt.py
+++ b/omc3/optics_measurements/rdt.py
@@ -185,7 +185,7 @@ def _fit_rdt_amplitudes(invariants, line_amp, plane, rdt):
     for i, bpm_rdt_data in enumerate(line_amp):
         popt, pcov = curve_fit(fitting, kick_data, bpm_rdt_data, p0=guess[i])
         amps[i] = popt[0]
-        err_amps[i] = np.sqrt(pcov)[0] if np.sqrt(pcov)[0] != np.Inf else 0. # if single file is used, the error is reported as Inf, which is then overwritten with 0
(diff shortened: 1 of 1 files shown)


mihofer:
Note: this error doesnt seem present in CRDT as there ODR is used to also take into account the error on the X (action)

omc3/optics_measurements/rdt.py
```diff
@@ -184,7 +184,8 @@ def fitting(x, f):
 
     for i, bpm_rdt_data in enumerate(line_amp):
         popt, pcov = curve_fit(fitting, kick_data, bpm_rdt_data, p0=guess[i])
-        amps[i], err_amps[i] = popt[0], np.sqrt(pcov)[0]
+        amps[i] = popt[0]
+        err_amps[i] = np.sqrt(pcov)[0] if np.sqrt(pcov)[0] != np.Inf else 0. # if single file is used, the error is reported as Inf, which is then overwritten with 0
```
fsoubelet, line 188: I would rather use `np.isfinite()` here as the comparison to and `Inf` value seems potentially dangerous.

fsoubelet: This would also catch `nan`s which isn't bad here.

fsoubelet (changes requested):

omc3/optics_measurements/rdt.py
fsoubelet: If we don't care about `nan`, let's go for `not np.isinf()` then.

fsoubelet (approved):
Great!

awegsche (approved):

fsoubelet:
I'm forcing the merge as the low coverage is irrelevant for this fix.

PR #353, opened 2021-11-17
Update MAD-X binaries from 5.06.00 to 5.07.00

### LINKED ISSUE #352: [Feature Request]: Update MAD-X Binaries
### Feature Description

While it is not primordial to instanltly and always be on the latest `MAD-X`, there has not been any update in a while.

### Possible Implementation

Downloading the releases from the MAD-X website and pushing the changes.

fsoubelet:
Closes #352 

Let's take this opportunity to see if the tests run properly with the new version, though there should be no hiccups.


mihofer (approved):
Long term, a switch to cpymad might be better, make things easier

fsoubelet:
While I fully agree @mihofer, that will be a massive amount of work and we clearly don't have time for it :(

mihofer:
> 
> 
> While I fully agree @mihofer, that will be a massive amount of work and we clearly don't have time for it :(

I quickly checked and massive might be an overstatement, if I didn't miss anything, one may "only" need to redo the `_run` function in the `madx_wrapper` https://github.com/pylhc/omc3/blob/a78e22cb83ed0f859e94865f33ffc6ca4175b0b2/omc3/madx_wrapper.py#L110

fsoubelet:
> > While I fully agree @mihofer, that will be a massive amount of work and we clearly don't have time for it :(
> 
> I quickly checked and massive might be an overstatement, if I didn't miss anything, one may "only" need to redo the `_run` function in the `madx_wrapper`
> 
> https://github.com/pylhc/omc3/blob/a78e22cb83ed0f859e94865f33ffc6ca4175b0b2/omc3/madx_wrapper.py#L110

That'd be the simple and ugly way to do it, the right way would be to make use of `cpymad`'s power by properly using it.

mihofer:
In the meantime, 5.08 is the latest MAD-X version.

PR #362, opened 2022-03-23
add 6800GeV to lhc uncertainties

awegsche:
as the title says.

the values are exactly the same as 6500GeV, our best guess so far


JoschD:
LGTM but I am not really the person to ask

fsoubelet:
Same with me. It would be good to wait for @tpersson on this

mihofer:
Since the tfs already has info in the header, might be interesting to add a `@ source` too for anyone to see where these numbers came from, if they are just the same as for 6.5 TeV, or come from Fidel/wise, or amalgamation of numbers from some meetings, etc.

awegsche:
added some info.

tpersson (approved):
The values look very reasonable to me but I haven't double-checked..

PR #366, opened 2022-04-11
Use the correct knobs for Run3 

tpersson:


omc3/model/constants.py
```diff
@@ -23,6 +23,7 @@
 
 GENERAL_MACROS = "general.macros.madx"
 LHC_MACROS = "lhc.macros.madx"
+LHC_MACROS_RUN3 = "lhc.macrosRun3.madx"
```
fsoubelet, line 26: Potentially `lhc.macros.run3.madx` for consistency

omc3/model/madx_macros/lhc.macrosRun3.madx
```diff
@@ -0,0 +1,33 @@
+/*
+* Macros for runI and runII of the LHC
```
fsoubelet, line 2: Would be nice to fix this info line too

fsoubelet (approved):
Small suggestions but nothing blocking, I'm happy merging like this too

awegsche (approved):

omc3/model/constants.py
tpersson: Thanks! fixed!

omc3/model/madx_macros/lhc.macrosRun3.madx
tpersson: Fixed!

tpersson:
> I am fully ok with this.
> 
> the `year == "2022"` will only help us for this year and the parenthesis in the `if` statement should be removed, if one wants to be nitpicky

Thanks! Removed the parenthesis. You are right that this only works for this year but was reluctant to do a Run3 solution since I am not sure how it will be the next year's + that this was the way it was done previously and it follows the logic of the acc-model repository. I suggest we leave it like this for now and then potentially change it next year when we have to update.

awegsche (approved):

PR #367, opened 2022-04-12
Release 0.2.5

fsoubelet:
Version update and changelog content for a patch release with the new LHC Run 3 knobs


fsoubelet:
I will bypass review as there is no code change and we want to publish this version today.

PR #370, opened 2022-04-25
add warning and error messages for phase output

awegsche:


fsoubelet (approved):
I am ok with this

omc3/optics_measurements/beta_from_phase.py
```diff
@@ -80,6 +80,12 @@ def n_bpm_method(meas_input, phase, plane, meas_and_mdl_tunes):
     Returns:
         `TfsDataFrame` containing betas and alfas from phase.
     """
+
+    if phases['MEAS'].count() < 3:
```
JoschD, line 84: why not raise an error?

omc3/optics_measurements/phase.py
```diff
@@ -66,6 +66,14 @@ def calculate(
                                                                              no_errors)
         dfs = free_dfs + drv_dfs
 
+    if phase_advances["MEAS"].count() < 3:
+        LOGGER.warning("Dear user, phase analysis yields less than 3 BPMs.")
```
JoschD, line 70: I'd write it into one single log with `\n` to separate lines. That makes it more obvious that it's one message. Also, `Dear user` might be too much

omc3/optics_measurements/beta_from_phase.py
```diff
@@ -398,6 +404,8 @@ def three_bpm_method(meas_input, phase, plane, meas_and_mdl_tunes):
     Returns:
         `TfsDataFrame` containing betas and alfas from phase.
     """
+    if phase["MEAS"].count() < 3:
```
JoschD, line 407: same as above. raise error?

JoschD (changes requested):

PR #375, opened 2022-05-08
Get Peak to Peak from SDDS files

JoschD:
After we had some problems yesterday with the pk2pk values in multiturn, I thought that might be helpful.
Strangely, I am not getting exactly the values from the logbook, e.g.

```python
from pathlib import Path

from omc3.scripts.check_peak_to_peak import peak_to_peak
from omc3.utils.logging_tools import get_logger, DEBUG

LOG = get_logger(__name__, level_console=DEBUG)


if __name__ == '__main__':
    base = Path('/user/slops/data/LHC_DATA/OP_DATA/FILL_DATA/7391/BPM')
    files = ['Beam2@BunchTurn@2018_10_30@12_54_03_710.sdds',  'Beam2@BunchTurn@2018_10_30@13_01_28_539.sdds', 'Beam2@BunchTurn@2018_10_30@13_03_01_002.sdds']
    peak_to_peak(files=[base/f for f in files])
```
```
Peak-to-Peak values for Beam2@BunchTurn@2018_10_30@13_01_28_539, Bunch  5:    X: 1.6512 mm    Y: 1.7183 mm
Peak-to-Peak values for Beam2@BunchTurn@2018_10_30@13_01_28_539, Bunch  5:    X: 1.6512 mm    Y: 1.7183 mm
Peak-to-Peak values for Beam2@BunchTurn@2018_10_30@13_01_28_539, Bunch  5:    X: 1.6512 mm    Y: 1.7183 mm
```
```
2018-10-30 12:54:03.710   - 5.0% 5.0% - 0.7265mm 0.7496mm 
2018-10-30 13:01:28.539  - 45.0% 35.0% - 1.635mm 1.6981mm 
2018-10-30 13:03:01.002  - 45.0% 35.0% - 1.6351mm 1.6748mm 
```
https://be-op-logbook.web.cern.ch/elogbook-server/GET/showEventInLogbook/3196358

Ideas are welcome! 
I am actually not sure, if such a script is needed in omc3 as it's probably one of those that everyone already has lying around anyway. Thank god I didn't write it from scratch.

### DIFF
--- a/omc3/scripts/check_peak_to_peak.py
+++ b/omc3/scripts/check_peak_to_peak.py
@@ -1,0 +1,199 @@
+"""
+Check Peak-to-Peak
+------------------
+
+Performs a quick check on the peak-to-peak of the given
+turn-by-turn files.
+"""
+import numbers
+import re
+from pathlib import Path
+from typing import Union, Sequence, Dict
+
+import numpy as np
+import pandas as pd
+import turn_by_turn as tbt
+from generic_parser import EntryPointParameters, entrypoint
+from turn_by_turn.utils import generate_average_tbtdata
+
+from omc3.definitions.constants import PLANES, UNIT_IN_METERS
+from omc3.harpy.constants import DEFOCUSSING_MONITORS
+from omc3.hole_in_one import HARPY_DEFAULTS
+from omc3.tbt_converter import _file_name_without_sdds
+from omc3.utils.logging_tools import get_logger
+
+LOG = get_logger(__name__)
+
+
+def get_params():
+    params = EntryPointParameters()
+    params.add_parameter(name="files", required=True, nargs='+',
+                         help="Files for analysis")
+    params.add_parameter(name="beam", type=int,
+                         help="LHC beam number.")
+    params.add_parameter(name="tbt_datatype",
+                         default=HARPY_DEFAULTS["tbt_datatype"],
+                         choices=list(tbt.io.DATA_READERS.keys()),
+                         help="Choose the datatype from which to import.")
+    params.add_parameter(name="unit", type=str, default="mm",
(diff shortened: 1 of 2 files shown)


omc3/scripts/check_peak_to_peak.py
```diff
@@ -0,0 +1,199 @@
...
+        beam = _get_beam(opt.beam, filename=name)
+        bpms = DEFOCUSSING_MONITORS[beam]
+        tbt_data = tbt.read_tbt(input_file, datatype=opt.tbt_datatype)
+        for bunch, positions in zip(tbt_data.bunch_ids, tbt_data.matrices):
+            LOG.debug(f"Bunch: {bunch}")
+            pk2pk = get_pk2pk(positions, bpms)
+            _log_pk2pk(pk2pk, name, bunch, opt.input_unit, opt.unit)
+
+        if tbt_data.nbunches > 1:
+            tbt_data_av = generate_average_tbtdata(tbt_data)
```
fsoubelet, line 105: Does Multiturn generate an average of the bunches too? If not, this could be why you observe a difference.

fsoubelet:
Otherwise, this is nice. Maybe we add a simple measurement as a test for this script?

JoschD:
missing 
- [ ] test
- [ ] include into docs
- [ ] add --bpms flag, to allow for 'all', 'lhc', 'lhcnonats',  and a given list of bpms

omc3/scripts/check_peak_to_peak.py
JoschD: I only generate an average if there is more than one bunch in the file. and then also additionally to the per bunch pk2pk. And for the tested files, there is only a single bunch. So no idea what the difference is

PR #376, opened 2022-05-19
Quick clean

JoschD:
Adds a script to clean columns from linfiles (or any tfs files, actually) via the `outliers.get_filter_mask` function.

### DIFF
--- a/omc3/scripts/linfile_clean.py
+++ b/omc3/scripts/linfile_clean.py
@@ -50,8 +50,6 @@ as well, as it only cleans on TUNE, not on NATTUNE.
 And it requires an accelerator instance.
 """
 import shutil
-from collections import OrderedDict
-from contextlib import suppress
 from numbers import Number
 from pathlib import Path
 from typing import Sequence, Union
@@ -59,14 +57,13 @@ from typing import Sequence, Union
 import pandas as pd
 import tfs
 from generic_parser.entrypoint_parser import (
-    save_options_to_config, entrypoint, EntryPointParameters
+    entrypoint, EntryPointParameters
 )
 
-from omc3.definitions import formats
 from omc3.definitions.formats import BACKUP_FILENAME
 from omc3.harpy.constants import COL_NAME
 from omc3.utils import logging_tools
-from omc3.utils.iotools import PathOrStr
+from omc3.utils.iotools import PathOrStr, save_config
 from omc3.utils.outliers import get_filter_mask
 
 LOG = logging_tools.get_logger(__name__)
@@ -110,7 +107,7 @@ def get_params():
 def main(opt):
     """Main function, to parse commandline input and separate restoration
     from cleaning."""
-    _save_options_to_config(opt)
+    save_config(Path('.'), opt, __file__)
     if opt.restore:
         restore_files(opt.files)
         return
@@ -183,7 +180,7 @@ def _filter_by_column(df: pd.DataFrame, column: str, limit: Number) -> pd.DataFr
         LOG.info(f"{column} not in current file. Skipping cleaning.")
         return df
 
-    good_bpms = get_filter_mask(df[column], limit)
+    good_bpms = get_filter_mask(data=df[column], limit=limit)
     n_good, n_total = sum(good_bpms), len(good_bpms)
     LOG.info(f"Cleaned {n_total-n_good:d} of {n_total:d} elements in {column} ({n_good:d} remaining).")
     return df.loc[good_bpms, :]
@@ -204,14 +201,6 @@ def _backup_file(file):
 
 def _get_backup_filename(file: Path, counter: int):
     return file.with_name(BACKUP_FILENAME.format(file.name, counter))
-
-
-# Helper -----------------------------------------------------------------------
-
-def _save_options_to_config(opt):
-    with suppress(IOError):
-        save_options_to_config(formats.get_config_filename(__file__),
-                               OrderedDict(sorted(opt.items())))
 
 
 # Script Mode ------------------------------------------------------------------
--- a/omc3/scripts/linfile_clean.py
+++ b/omc3/scripts/linfile_clean.py
@@ -141,15 +141,15 @@ def restore_files(files: Sequence[Union[Path, str]]):
 
 def _restore_file(file):
     counter = 1
-    backup_file = _get_backup_filename(file, counter)
+    backup_file = _get_backup_filepath(file, counter)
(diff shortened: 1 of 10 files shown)


omc3/scripts/linfile_clean.py
```diff
@@ -0,0 +1,209 @@
...
+        try:
+            _restore_file(file)
+        except IOError as e:
+            failed.append(str(e))
+
+    if len(failed):
+        all_errors = '\n'.join(failed)
+        if len(failed) == len(files):
+            raise IOError(f"Restoration of ALL files has failed\n{all_errors}")
+        raise IOError(f"Restoration of some files has failed, "
```
fsoubelet, line 136: We should specify in this error message that the printed files are those who failed to be restored.

omc3/scripts/linfile_clean.py
```diff
@@ -0,0 +1,209 @@
...
+        raise IOError(f"No backups found for file {file.name}")
+
+    # get last existing backup file
+    while backup_file.exists():
+        counter += 1
+        backup_file = _get_backup_filepath(file, counter)
+    backup_file = _get_backup_filepath(file, counter-1)
+
+    # restore found  file
+    LOG.info(f"Restoring last backup file {backup_file.name} to {file.name}.")
```
fsoubelet, line 155: I feel like maybe this should be a `debug` level logging statement

omc3/scripts/linfile_clean.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+# Clean ------------------------------------------------------------------------
+
+def clean_columns(files: Sequence[Union[Path, str]], columns: Sequence[str],
+                  limit: float = 0.0, backup: bool = True):
+    """ Clean the columns in the given files."""
+    for file in files:
+        file = Path(file)
+        LOG.info(f"Cleaning {file.name}.")
```
fsoubelet, line 166: Probably `debug` too. Could be a lot of logging output if a bunch of files are given

omc3/scripts/linfile_clean.py
```diff
@@ -0,0 +1,209 @@
...
+        if backup:
+            _backup_file(file)
+
+        tfs.write_tfs(file, df, save_index=COL_NAME)
+
+
+def _filter_by_column(df: pd.DataFrame, column: str, limit: Number) -> pd.DataFrame:
+    """Get the dataframe with all rows dropped filtered by the given column."""
+    if column not in df.columns:
+        LOG.info(f"{column} not in current file. Skipping cleaning.")
```
fsoubelet, line 180: Maybe `warning`?

omc3/scripts/linfile_clean.py
```diff
@@ -0,0 +1,209 @@
...
+        return df
+
+    good_bpms = get_filter_mask(data=df[column], limit=limit)
+    n_good, n_total = sum(good_bpms), len(good_bpms)
+    LOG.info(f"Cleaned {n_total-n_good:d} of {n_total:d} elements in {column} ({n_good:d} remaining).")
```
fsoubelet, line 185: I would also put `debug` level here

fsoubelet (commented):
This will be nice :)

Let's up the version number for when we merge this.

omc3/scripts/linfile_clean.py
JoschD: it does not write the filename, it writes the error messages. So it should be clear

omc3/scripts/linfile_clean.py
JoschD: that helps you to see, which files were restored though... I think that's important. These functions are not really useful anywhere else but in this script (for now). So I think `info` is good here (thought about it for quite a while)

omc3/scripts/linfile_clean.py
JoschD: yes. That's on purpose. It's not a function to be used, but a script.

omc3/scripts/linfile_clean.py
JoschD: No. Probably on purpose, if you want to clean `linx` and `liny` files in one go (I will call it twice from the GUI though). Then you can give it "TUNEX" and "TUNEY" as columns and it will just skipt it

omc3/scripts/linfile_clean.py
JoschD: same reason as above

omc3/scripts/linfile_clean.py
fsoubelet: That's my bad for reading too fast!

omc3/scripts/linfile_clean.py
fsoubelet: Let's think about this though. Given the amount of `info` level statements, if you give 5 files and 3 columns, you're already at 5 * 3 * 2 = 30 `info` level lines. This feels like polluting the output to me and drowning what is relevant information.

I'm ok with this specific line being `LOG.info` as long as the other ones are `debug` for instance.

JoschD: I'm all in favour of keeping the output minimal for functions. But this is a `script` that should be called from commandline and I want to see what has happened in here to my files. If on the other side, I activate debug loggig, THEN it will be super polluted, but all the downstream functions logging things.

fsoubelet: :man_shrugging:

fsoubelet (dismissed):

omc3/scripts/linfile_clean.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+@entrypoint(get_params(), strict=True)
+def main(opt):
+    """Main function, to parse commandline input and separate restoration
+    from cleaning."""
+    save_config(Path('.'), opt, __file__)
+    if opt.restore:
+        restore_files(opt.files)
+        return
```
mihofer, line 138: if you use this script as a function, it might be nice to return the files paths or something

JoschD: yeah. I thought about that, but then I realized, that the filepaths you return are just the file paths you put in...

mihofer: except for the files which couldnt be found/backup(p?)ed

JoschD: but that throws an error. so nothing is returned anyway

mihofer (approved):

fsoubelet (approved):

PR #377, opened 2022-05-22
fix .values and np.newaxis deprecation warnings

JoschD:


fsoubelet (approved):

fsoubelet:
Bit too many `to_numpy()` calls as they are in functions and calls apparently but fixing should be easy.

mihofer (approved):

fsoubelet (approved):

PR #379, opened 2022-05-29
Enhancement/2d ampdet kicks

JoschD:


omc3/plotting/plot_amplitude_detuning.py
```diff
@@ -329,23 +385,325 @@ def _format_axes(ax, limits, labels):
...
+    # ax.plot_surface(xm, ym, f_upp(xm, ym), color=color, alpha=0.3)
+
+    # Plot the full surface
+    x = np.linspace(0, xymax[0], 2)
+    y = np.linspace(0, xymax[1], 2)
+    plot_cube(ax, x, y, f_low, f_upp, color=color, alpha=0.2, rcount=2, ccount=2)
+    return color
+
+
+def plot_cube(ax: Axes, x: ArrayLike, y: ArrayLike, f_low: callable, f_upp: callable, **kwargs):
```
fsoubelet, line 577: Nice!

omc3/tune_analysis/fitting_tools.py
```diff
@@ -44,11 +51,131 @@ def do_odr(x, y, xerr, yerr, order):
     Returns: Odr fit. Betas order is index = coefficient of same order.
     """
     LOG.debug("Starting ODR fit.")
-    fit_np = np.polyfit(x, y, order)[::-1]  # using polyfit as starting parameters
+
+    # Poly-Fit for starting point ---
+    fit_np = np.polyfit(x, y, order)[::-1]  # polyfit order is inversed to poly_func
```
fsoubelet, line 56: If this isn't too much work, maybe we could use `Polynomial.fit` as `numpy` recommends for new code: https://numpy.org/doc/stable/reference/routines.polynomials.html

omc3/tune_analysis/kick_file_modifiers.py
```diff
@@ -217,42 +247,134 @@ def get_ampdet_data(kickac_df, action_plane, tune_plane, corrected=False):
...
     df.index = pd.Index([CERNDatetime.from_cern_utc_string(i) for i in df.index], dtype=object)
     return df
 
 
-def write_timed_dataframe(path, df):
+def write_timed_dataframe(path: Union[str, Path], df: tfs.TfsDataFrame):
     df = df.copy()
     df.index = pd.Index([i.strftime(get_cern_time_format()) for i in df.index], dtype=str)
-    tfs.write(str(path), df, save_index=get_time_col())
+    tfs.write(path, df, save_index=get_time_col(), headerswidth=max(len(k) for k in df.headers.keys()))
```
fsoubelet, line 282: I'm thinking maybe we should have this be the default for `tfs-pandas` writing at some point if `headerswidth` is not given?

omc3/utils/logging_tools.py
```diff
@@ -208,16 +209,21 @@ def get_log(self):
...
-    printer(f'  Beta: {odr_out.beta}')
-    printer(f'  Beta Std Error: {odr_out.sd_beta}')
-    printer(f'  Beta Covariance: {odr_out.cov_beta}')
+    old_opts = np.get_printoptions()
+    np.set_printoptions(precision=2)
+    odr_str = ('\nODR-Summary:\n'
+               f'Beta:\n {odr_out.beta}\n'
+               f'Beta Std Error:\n {odr_out.sd_beta}\n'
+               f'Beta Covariance:\n {odr_out.cov_beta}\n'
+               )
```
fsoubelet, line 218: No indentation here but indentation below?

tests/unit/test_amplitude_detuning.py
```diff
@@ -1,90 +1,98 @@
...
 
 from omc3.amplitude_detuning_analysis import analyse_with_bbq_corrections
-from omc3.plotting.plot_amplitude_detuning import main as pltampdet
 from omc3.plotting.plot_bbq import main as pltbbq
 
-# Forcing non-interactive Agg backend so rendering is done similarly across platforms during tests
-matplotlib.use("Agg")
-
 
 @pytest.mark.basic
```
fsoubelet, line 10: Marked as `basic` but calls a fixture marked as `extended`. Is this wanted?

omc3/tune_analysis/fitting_tools.py
fsoubelet: Order of returned coefficients seems opposite though. So the [::-1] from here would go away

tests/unit/test_amplitude_detuning.py
JoschD: yes. the run with the filter method "outliers" is the basic run (i.e. it tests most of the functionality of the detuning analysis already and if this fails there is something seriously broken). Then the same function is called with different bbq filtering methods, so only that part is different, therefore they are extended. 
If these were proper unit tests, that would look very different ...

omc3/tune_analysis/kick_file_modifiers.py
JoschD: possibly, yeah.

fsoubelet (approved):

PR #384, opened 2022-06-21
LHC macros call fix

fsoubelet:
Currently the LHC model creator will call ats macros (tune matching for instance) only if the year is exactly "2018". Similarly, it calls the run3 macros file only if the year is exactly "2022".

This quick change checks that the year is `>=` to these values (with `int` conversion) and not strictly equal.
It also corrects the run 3 macros to use the proper knobs (`op` or `sq` depending on optics).

I have updated the version number to make this a patch release.

### DIFF
--- a/omc3/model/accelerators/lhc.py
+++ b/omc3/model/accelerators/lhc.py
@@ -334,7 +334,7 @@ class Lhc(Accelerator):
             f"call, file = '{self.model_dir / MACROS_DIR / GENERAL_MACROS}';\n"
             f"call, file = '{self.model_dir / MACROS_DIR / LHC_MACROS}';\n"
             )
-        if self._uses_run3_macros():  # include the Run 3 macros (fsoubelet, 2022)
+        if self._uses_run3_macros():
             madx_script += (
                 f"call, file = '{self.model_dir / MACROS_DIR / LHC_MACROS_RUN3}';\n"
             )
--- a/omc3/model/accelerators/lhc.py
+++ b/omc3/model/accelerators/lhc.py
@@ -330,7 +330,7 @@ class Lhc(Accelerator):
             f"call, file = '{self.model_dir / MACROS_DIR / GENERAL_MACROS}';\n"
             f"call, file = '{self.model_dir / MACROS_DIR / LHC_MACROS}';\n"
             )
-        if int(self.year) >= 2022:  # include the Run 3 macros (fsoubelet, 2022)
+        if self.year.lower().startswith("hllhc") or int(self.year) >= 2022:  # include the Run 3 macros (fsoubelet, 2022)
             madx_script += (
                 f"call, file = '{self.model_dir / MACROS_DIR / LHC_MACROS_RUN3}';\n"
             )
@@ -365,7 +365,7 @@ class Lhc(Accelerator):
         if high_beta:
             madx_script += "exec, high_beta_matcher();\n"
 
-        if int(self.year) >= 2018:  # starting there we use ATS optics and macros (fsoubelet, 2022)
+        if self.year.lower().startswith("hllhc") or int(self.year) >= 2018:  # starting there we use ATS optics and macros (fsoubelet, 2022)
             madx_script += f"exec, match_tunes_ats({self.nat_tunes[0]}, {self.nat_tunes[1]}, {self.beam});\n"
         else:
             madx_script += f"exec, match_tunes{ats_suffix}({self.nat_tunes[0]}, {self.nat_tunes[1]}, {self.beam});\n"
--- a/omc3/model/accelerators/lhc.py
+++ b/omc3/model/accelerators/lhc.py
@@ -330,7 +330,7 @@ class Lhc(Accelerator):
             f"call, file = '{self.model_dir / MACROS_DIR / GENERAL_MACROS}';\n"
             f"call, file = '{self.model_dir / MACROS_DIR / LHC_MACROS}';\n"
             )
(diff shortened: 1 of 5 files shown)


JoschD:
You should check for `self.year.startswith("hllhc")` or so (which is the reason, why this is not an `int` in the first place)

JoschD:
also, add to changelog:

```
#### 2022-06-21 - v0.4.1 - _jdilly_, _fesoubel_

- Fixed:
  - LHC ATS macros called for all years after 2018
  - LHC RunIII macros called for all years after 2022
  - Getting new BBQ data ended in a key-error.
  - Better KeyError Message for Exciter-BPM not found.
```

JoschD (changes requested):
see above messages

fsoubelet:
Good catch, done!

awegsche:
Hi,

I didn't go through the code in detail yet. But in the comments and description there is mention of ATS macros.

As of the 2021 optics, ATS knobs **are not to be used** manually, but the `_op` ones must be used. Handling of ATS factors and on/off is done internally by `acc-models`.

If this is just a naming issue, please consider renaming the `macro`s to avoid confusion.

JoschD:
> Hi,
> 
> I didn't go through the code in detail yet. But in the comments and description there is mention of ATS macros.
> 
> As of the 2021 optics, ATS knobs **are not to be used** manually, but the `_op` ones must be used. Handling of ATS factors and on/off is done internally by `acc-models`.
> 
> If this is just a naming issue, please consider renaming the `macro`s to avoid confusion.

Why don't you go through the code in detail then and tell us how it is done correctly?

awegsche:
The main reason for my comment is to prevent this to be merged too quickly. 

And then a bit of courtesy I guess.

omc3/model/madx_macros/lhc.macros.run3.madx
```diff
@@ -1,23 +1,43 @@
...
 
 
+/*
+* Performs the matching of the LHC tunes.
+* This is adapted for Run 3 to use the *_op knobs
+* @param nqx: The horizontal tune to match to.
+* @param nqy: The vertical tune to match to.
+* @param beam_number: The beam to use, either 1 or 2.
+*/
+match_tunes(nqx, nqy, beam_number): macro = {
```
awegsche, line 16: this macro is actually correct, it should be called (and **not** the `match_tunes_ats` one)

omc3/model/accelerators/lhc.py
```diff
@@ -384,6 +388,19 @@ def get_update_correction_script(self, outpath: Path, corr_file: Path) -> str:
         )
         return madx_script
 
+    def _uses_ats_knobs(self) -> bool:
+        """Returns wether the ATS knobs and macros should be used, based on the instance's year."""
+        try:
+            return int(self.year) >= 2018  # self.year is always a string
```
awegsche, line 394: and here there should be a `==`, since using ATS knobs is an outlier from OP's point of view

omc3/model/madx_macros/lhc.macros.run3.madx
```diff
@@ -1,23 +1,43 @@
...
+
 /*
 * Performs the matching of the LHC tunes, adapted to ATS optics.
+* This is adapted for Run 3 to use the *_sq knobs
 * @param qx: The horizontal tune to match to.
 * @param qy: The vertical tune to match to.
-* @param beam_number: The beam to use either 1 or 2.
+* @param beam_number: The beam to use, either 1 or 2.
 */
 match_tunes_ats(nqx, nqy, beam_number): macro = {
```
awegsche, line 34: matching on `_sq` knobs is considered hacky (I guess) and could be kept only for legacy reasons / to do some hacks in the control room  (like, e.g. when the `acc-models` optics files are missing the definitions of `_op` knobs, or mix up `:=` and `=`, etc)

JoschD: I think we can leave the macro in, but not use it and write this as a comment into the .madx file directly as a warning for future users.

omc3/model/accelerators/lhc.py
JoschD: that's what I feared. Easy fix :). Thx!

omc3/model/madx_macros/lhc.macros.run3.madx
fsoubelet: I thought we used the ATS macros in 2021 too?

omc3/model/madx_macros/lhc.macros.run3.madx
fsoubelet: I will leave them in, with a comment explaining things

fsoubelet:
Here's the current state of this:

- Run 3 macros are called if the year is >= 2022
- ATS knobs versions of the macros (`match_tunes_ats` and `coupling_knob_ats`) are called for years 2018 to 2021 (inclusive) or any time `--ats` is provided to the model creator
- The run3 macros file is updated to use `_op` knobs by default (`match_tunes` and `coupling_knob`), but leaves a `match_tunes_ats` that uses the `_sq` knobs in case we specifically want it (that would be by providing `--ats`)

@awegsche could you please confirm this is correct? @JoschD and I would like to have this patch release done before tonight's MD.

JoschD (approved):
LGTM

awegsche:
The knobs look fine

awegsche (approved):
verified the knobs, this is good to be shipped

Issue #383, opened 2022-06-13
[Feature Request]: Add higher orders of multipoles for RDTs


Mael-Le-Garrec:
### Feature Description

The current RDT implementation is limited to sextupoles and octupoles, as the f-terms are hardcoded at the beginning of [rdt.py](https://github.com/pylhc/omc3/blob/master/omc3/optics_measurements/rdt.py).

To study higher order RDTs, like $f_{1004}$ generated by decapoles, it is needed to add the f-term lines to the dict.

For those RDT calculations being possible, it is also needed to calculate the amplitude of the lines in harpy. The lines can be found in the `RESONANCES` dict in [frequency.py](https://github.com/pylhc/omc3/blob/master/omc3/harpy/frequency.py)

### Possible Implementation

Either add a bunch of f-terms in `rdt.py` and the associated amplitude lines in `frequency.py`
Or write more general functions that would generate those terms/lines given a multipole order.

Mael-Le-Garrec:
Fixed by PR https://github.com/pylhc/omc3/pull/385

PR #385, opened 2022-06-21
Higher RDT multipole order calculation

Mael-Le-Garrec:
Issue #383 

The current RDT implementation is limited to sextupoles and octupoles, as the f-terms are hardcoded at the beginning of [rdt.py](https://github.com/pylhc/omc3/blob/master/omc3/optics_measurements/rdt.py).

To study higher order RDTs, like $f_{1004}$ generated by decapoles, it is needed to add the f-term lines to the dict.

For those RDT calculations being possible, it is also needed to calculate the amplitude of the lines in harpy. The lines can be found in the `RESONANCES` dict in [frequency.py](https://github.com/pylhc/omc3/blob/master/omc3/harpy/frequency.py)


JoschD (dismissed):
LGTM as far as I understand the RDT sorting. Would be cool to get some feedback from @lmalina . But also maybe @mihofer could give it a quick glance

omc3/optics_measurements/rdt.py
```diff
@@ -59,14 +83,22 @@ def calculate(measure_input, input_files, tunes, phases, invariants, header):
         for_rdts = _best_90_degree_phases(meas_input, bpm_names, phases, tunes, plane)
         LOGGER.info(f"Average phase advance between BPM pairs: {for_rdts.loc[:,'MEAS'].mean()}")
         for rdt in SINGLE_PLANE_RDTS[plane]:
-            df = _process_rdt(meas_input, input_files, for_rdts, invariants, plane, rdt)
+            try:
+                df = _process_rdt(meas_input, input_files, for_rdts, invariants, plane, rdt)
+            except ValueError as e:  # catch the AMP line not being found in the lin file
+                LOGGER.warning(f"RDT calculation failed for {jklm2str(*rdt)}: {str(e)}")
+                continue
             write(df, add_freq_to_header(header, plane, rdt), meas_input, plane, rdt)
```
fsoubelet, line 91: Shouldn't the write line be in the try block? As I see it, a failure in `_process_rdt` will log a warning but still try to write to disk, which would fail.

omc3/optics_measurements/rdt.py
```diff
@@ -59,14 +83,22 @@ def calculate(measure_input, input_files, tunes, phases, invariants, header):
...
         for_rdts = _best_90_degree_phases(meas_input, bpm_names, phases, tunes, plane)
         LOGGER.info(f"Average phase advance between BPM pairs: {for_rdts.loc[:, 'MEAS'].mean()}")
         for rdt in DOUBLE_PLANE_RDTS[plane]:
-            df = _process_rdt(meas_input, input_files, for_rdts, invariants, plane, rdt)
+            try:
+                df = _process_rdt(meas_input, input_files, for_rdts, invariants, plane, rdt)
+            except ValueError as e:
+                LOGGER.warning(f"RDT calculation failed for {jklm2str(*rdt)}: {str(e)}")
+                continue
             write(df, add_freq_to_header(header, plane, rdt), meas_input, plane, rdt)
```
fsoubelet, line 102: Same thing here.

omc3/optics_measurements/rdt.py
```diff
@@ -209,7 +249,11 @@ def get_line_sign_and_suffix(line, input_files, plane):
         return 1, suffix
     if f"AMP{conj_suffix}" in input_files[plane][0].columns:
         return -1, conj_suffix
-    raise ValueError(f"No data for line {line} in plane {plane}")
+
+    # The specified AMP column hasn't been found in the lin file
+    msg = (f"The column AMP{suffix} has not been found in the lin{plane.lower()} file. "
+            "Consider re-running the frequency analysis with a higher order resonance term")
```
fsoubelet, line 255: What flag for `harpy` would that correspond to?

fsoubelet (commented):
Few remarks. It would be nice to add a few tests for the new calculated orders, similar to the existing ones.

omc3/optics_measurements/rdt.py
Mael-Le-Garrec: good catch, I just moved them

omc3/optics_measurements/rdt.py
Mael-Le-Garrec: There are no flags, it is referencing this line `RESONANCES = get_resonance_lines(5)`
I'm not sure how to implement that with a flag at the moment

JoschD: Yes. It's a bit more involved as it should be a common parameter in both harpy and measure optics. And I think if not given it should default to something lower though. 
As mael was quite busy we said we can introduce this in two steps (i.e. two PRs) but as we want. Important here would be to make sure the principle is correct.

JoschD: (Also, the resonances dict could then be just a function instead of a global parameter)

fsoubelet: I agree this can ship as is for now. I don't think it matters too much to make it a lower value by default as it only kicks in if `--rdt` is provided to `measure_optics`.

JoschD: that is the problem: it also kicks in in harpy. So the lin-files get bigger. That's also why you need this flag in both

omc3/optics_measurements/rdt.py
```diff
@@ -18,26 +18,50 @@
...
+            single_plane['X'].append((j,k,l,m))
+        elif j+k == 0 and l != 0:  # same, but for the Qy line
+            single_plane['Y'].append((j,k,l,m))
+        elif j == 0:  # The RDT can only be seen on the vertical plane and uses both Qx and Qy
+            double_plane['Y'].append((j,k,l,m))
+        elif l == 0: # same, but for the horizontal plane
+            double_plane['X'].append((j,k,l,m))
+        else:  # The RDT can be seen on both planes and is a multiple of both Qx and Qy
+            double_plane['X'].append((j,k,l,m))
+            double_plane['Y'].append((j,k,l,m))
```
mihofer, line 59: might be nice to give one or two example RDTs for each if statement, such as normal sextupole f3000 and normal octupole f4000 for the single X, and the skew counterparts in the Y

mihofer (dismissed):
Good for me. some minor things, but not breaking.

omc3/optics_measurements/rdt.py
Mael-Le-Garrec: I just added arguments for both harpy and the optics:
* harpy: --resonances
* optics: --rdt_magnet_order, to avoid confusing it with the order of the RDT

omc3/optics_measurements/rdt.py
Mael-Le-Garrec: Done

fsoubelet (changes requested):
Fails on windows with the temporary directory thingy, which is an issue I think we had in the past.

Can you use the [tmp_path](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-fixture) fixture instead? Instead of `_test_directory` use `tmp_path` in the test arguments and body.

See for example here: https://github.com/pylhc/omc3/blob/24f0cb0797ed784ec610591a01f36f111dce414b/tests/unit/test_plot_optics_measurements.py#L17=

fsoubelet (dismissed):

fsoubelet (dismissed):
I've taken liberty to bump the version and add a few lines to the changelog so this can be the next release. @Mael-Le-Garrec let me know if anything there looks off.

omc3/hole_in_one.py
```diff
@@ -204,6 +204,11 @@ def hole_in_one_entrypoint(opt, rest):
         Flags: **--window**
         Choices: ``('rectangle', 'welch', 'triangle', 'hann', 'hamming', 'nuttal3', 'nuttal4')``
         Default: ``hann``
+      - **resonances** *(int)*: Maximum magnet order of resonance lines to calculate.
+
+        Flags: **--resonances**
+        Choices: ``(n>=2)``
+        Default: ``4``
```
mihofer, line 211: maybe add which notation is used (European b1 = dipole, American/MAD-X convention b0 = dipole)

tests/inputs/models/2022_inj_b1_acd/job.twiss.madx
```diff
@@ -0,0 +1,45 @@
...
+
+use, period = LHCB1;
+
+option, echo;
+
+exec, match_tunes(0.28, 0.31, 1);
+
+exec, do_twiss_monitors(LHCB1, "/afs/cern.ch/work/m/mlegarre/public/jupyter/resonance_driving_terms/measurements/2022-06-24/Analysis/Model/twiss.dat", 0.0);
+exec, do_twiss_elements(LHCB1, "/afs/cern.ch/work/m/mlegarre/public/jupyter/resonance_driving_terms/measurements/2022-06-24/Analysis/Model/twiss_elements.dat", 0.0);
+
```
mihofer, line 40: maybe use a relative path to current dir instead

omc3/hole_in_one.py
mihofer: and Choices: ( 6 >= n >= 2)

omc3/hole_in_one.py
```diff
@@ -497,11 +510,18 @@ def harpy_params():
...
+                        help="Maximum magnet order of resonance lines to calculate.")
     return params
 
 
 def _optics_entrypoint(params):
-    return EntryPoint(optics_params(), strict=False).parse(params)
+    options, rest = EntryPoint(optics_params(), strict=False).parse(params)
+    
+    if options.rdt_magnet_order < 2 and 'rdt' in options.nonlinear:
+        raise AttributeError("The minimum magnet order for RDT calculation is 2.")
```
mihofer, line 522: shouldn't those also check the upper bound?

omc3/optics_measurements/rdt.py
```diff
@@ -18,26 +18,50 @@
...
+        elif l == 0: # same, but for the horizontal plane
+            double_plane['X'].append((j,k,l,m))  # e.g. f1001, f1002
+        else:  # The RDT can be seen on both planes and is a multiple of both Qx and Qy
+            double_plane['X'].append((j,k,l,m))  # e.g. f1020, f1120
+            double_plane['Y'].append((j,k,l,m))
+
+    return single_plane, double_plane
+
+# TODO: add the order in the arguments parser
+#SINGLE_PLANE_RDTS, DOUBLE_PLANE_RDTS = _generate_plane_rdts(5)
```
mihofer, line 64: remove

omc3/hole_in_one.py
Mael-Le-Garrec: Voilà

tests/inputs/models/2022_inj_b1_acd/job.twiss.madx
Mael-Le-Garrec: I simply removed the useless files of the model

omc3/hole_in_one.py
Mael-Le-Garrec: Voilà

omc3/optics_measurements/rdt.py
Mael-Le-Garrec: Voilà

mihofer (approved):

fsoubelet (approved):

PR #386, opened 2022-06-30
Proper use of vectorize

fsoubelet:
The `np.vectorize` function returns a callable which has to be assigned, and previously was not. This is a simple fix.


Mael-Le-Garrec (approved):

JoschD (approved):

JoschD:
Looks good, doesn't work ^^

fsoubelet:
Yes, I'll have to look into it. If the performance gains are not so impressive I might just drop these two lines altogether.

PR #389, opened 2022-09-12
integrate knob extractor in omc3

awegsche:
_AND rename `--extract` to `--time` as result of vote_

This moves the python3 script `knob_extractor` from [Beta-Beat.src](https://github.com/pylhc/Beta-Beat.src) to omc3.

This is also a preparing step for the upcoming `model_creator` overhaul.


Mael-Le-Garrec:
Do you have tests?
It'd be nice to at least check the parser

awegsche:
> Do you have tests? It'd be nice to at least check the parser

Does python do docstring tests? We could transform the usage-examples into tests.

awegsche:
> Do you have tests? It'd be nice to at least check the parser

Does python do docstring tests? We could transform the usage-examples into tests.

awegsche:
there is a first test

Mael-Le-Garrec:
There are many `TODO` in the code, can you review them quickly and add the missing examples and the necessary checks?

There still is a comment about the non UTC date formats in the tests, can you take a look?

JoschD:
Feature implemented by #394

PR #394, opened 2022-09-19
Knob extractor plus

JoschD:
Alternative suggestion to the knob-extractor implementation.


omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+from dateutil.relativedelta import relativedelta
+
+from generic_parser import EntryPointParameters, entrypoint
+from omc3.utils.iotools import PathOrStrOrDataFrame, PathOrStr
+from omc3.utils.logging_tools import get_logger
+from omc3.utils.mock import cern_network_import
+
+LOGGER = get_logger(__name__)
+
+pytimber = cern_network_import("pytimber")
```
awegsche, line 81: If I understand correctly, this still tries to import `pytimber` directly. In that case, this should be moved down

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+    madx: str
+    lsa: str
+    scaling: float  # is usually +-1, i.e. takes care of sign-conventions
+    value: float = None
+
+    def get_madx(self):
+        if self.value is None:
+            return f"! {self.madx} : No Value extracted"
+        return f"{self.madx} := {self.value * self.scaling};"
+
```
awegsche, line 217: I like `structs`. but at this point this is just bloat

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+        Dictionary with LSA names (but with colon instead of /) as
+        keys and KnobEntries (without values) as value.
+
+    """
+    df.columns = df.columns.astype(str).str.lower()
+    df = df[['lsa', 'madx', 'scaling']].set_index("lsa", drop=False)
+    return {
+        lsa2name(r[0]): KnobEntry(**r[1].to_dict()) for r in df.iterrows()
+    }
+
```
awegsche, line 387: Again, this is nice to have, no question. But there is no intention to use such a functionality

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+    # following ISO-8601 for time durations
+    char_map = dict(
+        s='seconds', m='minutes', h='hours',
+        d='days', w='weeks', M='months', Y="years",
+    )
+
+    # add all deltas
+    time_parts = {char_map[delta[1]]: sign * int(delta[0]) for delta in all_deltas}
+    time = time + relativedelta(**time_parts)
+
```
awegsche, line 444: this is a bit hard to understand. this can arguably be left as-is, but it _should_ be either less confusing or have some comments explaining what is going on.

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+# Other tools ------------------------------------------------------------------
+
+def lsa2name(lsa_name: str) -> str:
+    return lsa_name.replace("/", ":")
+
+
+def name2lsa(name: str) -> str:
+    return name.replace(":", "/")
+
```
awegsche, line 456: nice to have but relevant?

Mael-Le-Garrec (changes requested):
Could you add a test for the parser?
Like testing non existing arguments, wrong types, etc.

Mael-Le-Garrec (changes requested):
The `README` could be updated to add the script

omc3/knob_extractor.py
Mael-Le-Garrec: I agree, it took me some time to get it.
Once you know what it does it's fine, but a comment would make it obvious at the first sight

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+            ),
+        ),
+    )
+
+
+@dataclass
+class KnobEntry:
+    madx: str
+    lsa: str
+    scaling: float  # is usually +-1, i.e. takes care of sign-conventions
```
fsoubelet, line 210: ±

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+    argument_parser_args=dict(
+        epilog=USAGE_EXAMPLES,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        prog="Knob Extraction Tool."
+    )
+)
+def main(opt) -> Optional[KnobsDict]:
+    """ Main knob extracting function. """
+    if not opt.time and not opt.state:
+        raise ValueError("No functionality selected. Set either `time` or `state`.")
```
fsoubelet, line 233: Maybe we use `ArgumentError`?

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+
+# Knobs Dict -------------------------------------------------------------------
+
+def _get_knobs_def_file(user_defined: Optional[Union[Path, str]] = None) -> Path:
+    """ Check which knobs-definition file is appropriate to take. """
+    if user_defined is not None:
+        LOGGER.info(f"Using user defined knobs.txt: '{user_defined}")
+        return Path(user_defined)
+
+    if KNOBS_TXT_MDLDIR.is_file():
```
fsoubelet, line 334: See, here the variable with `MDLDIR` makes me think it's a directory, and we check for `.is_file()`. No me gusta :/

omc3/knob_extractor.py
JoschD: moved up?

omc3/knob_extractor.py
JoschD: yes, because if this ever changes, you don't need to change it everywhere in the code. 
Also it is kind of documenting what it is doing in the code.

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+    """ Parse time from given string. """
+    # Now? ---
+    if time_str.lower() == "now":
+        return datetime.now()
+
+    # ISOFormat? ---
+    try:
+        return datetime.fromisoformat(time_str)
+    except (TypeError, ValueError):
+        pass
```
fsoubelet, line 419: Maybe a `LOGGER.debug("Could not parse time string as ISO format")` before we `pass` would be nice.

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+
+    # Timestamp? ---
+    try:
+        return datetime.fromtimestamp(int(time_str))
+    except (TypeError, ValueError):
+        pass
```
fsoubelet, line 425: Idem here

omc3/knob_extractor.py
awegsche: maybe we can add a lazy initialisation to the mocking ` pytimber.LoggingDB()` function

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+        # only print the state of the StateTracker - the MetaState!
+        # I still don't know what this does,
+        # because I only get back that the variable does not exist. (jdilly)
+        state = ldb.get("LhcStateTracker:State", time)  # do first to have output together
+        LOGGER.info("---- STATE ------------------------------------")
+        LOGGER.info(state)
+        return None
+    
+    knobs_dict = _parse_knobs_defintions(opt.knob_definitions)
+    knobs_extract = _extract(ldb, knobs_dict, opt.knobs, time)
```
fsoubelet, line 248: Am I correct in understanding that we can provide **both** the `--knobs` and `--knobs_definitions`? I feel this should be made a little bit clearer in the documentation / help strings.

omc3/knob_extractor.py
JoschD: I don't understand. This should be handled like a normal import, no?

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+    _write_knobsfile(opt.output, knobs_extract, time)
+    return knobs_extract
+
+
+def _extract(ldb, knobs_dict: KnobsDict, knob_categories: Sequence[str], time: datetime) -> KnobsDict:
+    """
+    Main function to gather data from  the state-tracker.
+
+    Args:
+        ldb: The pytimber database.
```
fsoubelet, line 258: I know the type hint would raise an error in the definition, but we can add `(pytimber.LoggingDB)` here in the docstring.

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+            knobs[knob] = knobs_dict[knob]
+
+            LOGGER.info(f"Looking for {knob:<34s} ")
+            knobkey = f"LhcStateTracker:{knob}:target"
+            knobvalue = ldb.get(knobkey, time.timestamp())  # use timestamp to preserve timezone info
+            if knobkey not in knobvalue:
+                LOGGER.warning(f"No value for {knob} found!")
+                continue
+
+            LOGGER.debug(f"Some value for {knob} extracted.")
```
fsoubelet, line 284: `LOGGER.debug(f"Extracted the value of knob '{knob}'")`

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+    LOGGER.info("---- EXTRACTING KNOBS -------------------------")
+    LOGGER.info(f"extracting knobs for {time}")
+    knobs = {}
+
+    LOGGER.info("---- KNOBS ------------------------------------")
+    for category in knob_categories:
+        for knob in KNOB_CATEGORIES.get(category, [category]):
+            knobs[knob] = knobs_dict[knob]
+
+            LOGGER.info(f"Looking for {knob:<34s} ")
```
fsoubelet, line 277: I feel like this should be a `.debug` level.

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+        knob_categories (Sequence[str]): Knob Categories or Knob-Names to extract.
+        time (datetime): The time, when to extract.
+
+    Returns:
+        Dict[str, KnobsDict]: Contains all the extracted knobs, grouped by categories.
+        When extraction was not possible, the value attribute of the respective KnobEntry is still None
+
+    """
+    LOGGER.info("---- EXTRACTING KNOBS -------------------------")
+    LOGGER.info(f"extracting knobs for {time}")
```
fsoubelet, line 269: I'd rather have just one statement here. `LOGGER.info(f"----- Extracting knobs for time {time} -----")`

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+    knobs = {}
+
+    LOGGER.info("---- KNOBS ------------------------------------")
```
fsoubelet, line 272: Don't think this one is necessary.

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+            knobkey = f"LhcStateTracker:{knob}:target"
+            knobvalue = ldb.get(knobkey, time.timestamp())  # use timestamp to preserve timezone info
+            if knobkey not in knobvalue:
+                LOGGER.warning(f"No value for {knob} found!")
+                continue
+
+            LOGGER.debug(f"Some value for {knob} extracted.")
+            timestamps, values = knobvalue[knobkey]
+            if len(values) == 0:
+                LOGGER.warning(f"No value for {knob} found for given time!")
```
fsoubelet, line 287: `LOGGER.warning(f"No value for {knob} found!")` for consistency with above. Or the other way around, but let's have both be the same.

omc3/knob_extractor.py
```diff
@@ -0,0 +1,459 @@
...
+                continue
+
+            value = values[-1]
+            if not math.isfinite(value):
+                LOGGER.warning(f"Value for {knob} is not a number or infinite!")
+                continue
+
+            LOGGER.info(f"Knob value for {knob} extracted: {value} (unscaled)")
```
fsoubelet, line 295: Redundant with line 284 (`LOGGER.debug(f"Some value for {knob} extracted.")`). I would rather have just this one then, since it's the same but with the value.

tests/unit/test_knob_extractor.py
```diff
@@ -0,0 +1,221 @@
...
+    return d, txt
+
+
+@pytest.fixture()
+def knob_definitions():
+    return INPUTS / "knob_definitions.txt"
+
+
+@pytest.fixture()
+def saved_knob_txt():
```
fsoubelet, line 220: Give me a type hiint here at east that is also returns the timestamp. Not clear from the fixture name

tests/unit/test_knob_extractor.py
```diff
@@ -0,0 +1,221 @@
...
+        if not line or line.startswith("!"):
+            continue
+
+        match = pattern.match(line)
+        d[match.group(1)] = float(match.group(2))
+    return d, txt
+
+
+@pytest.fixture()
+def knob_definitions():
```
fsoubelet, line 215: `-> Path:`

tests/unit/test_knob_extractor.py
```diff
@@ -0,0 +1,221 @@
...
+
+    assert len(parsed_saved) == len(parsed_output)
+    for key in parsed_output.keys():
+        assert parsed_output[key] == parsed_saved[key]
+
+
+# Helper -----------------------------------------------------------------------
+
+
+def parse_output_file(file_path):
```
fsoubelet, line 200: type hints

tests/unit/test_knob_extractor.py
```diff
@@ -0,0 +1,221 @@
...
+            _parse_time("2022-06-26T03:00+02:00")
+            ==
+            datetime(2022, 6, 26, 3, 0, 0, tzinfo=timezone(timedelta(seconds=7200)))
+    )
+
+# CERN Tests -------------------------------------------------------------------
+
+
+@pytest.mark.cern_network
+def test_real(tmp_path, knob_definitions, saved_knob_txt):
```
fsoubelet, line 184: maybe rename to `test_extractor_in_cern_network` or something more explicit

fsoubelet (commented):
I'm nitpicky, I know ;)

omc3/knob_extractor.py
fsoubelet: What's the issue with organised data structures?

omc3/knob_extractor.py
fsoubelet: What do you mean? It's convenient for anything that stays in the Python process and doesn't want to go through command line level.

omc3/knob_extractor.py
fsoubelet: Agreed, the comprehension is succint but the comment line above could detail what each of the indices used represent. Don't want to scroll the file to look for what `delta[1]` is etc.

omc3/knob_extractor.py
Mael-Le-Garrec: @fsoubelet, what Andreas meant to say is that pytimber starts pulling java libraries when imported (or when LoggingDB called?)
This means you'd have to wait quite a long time, only to check if your arguments make sense for example

omc3/knob_extractor.py
JoschD: this is outdated anyway. sorry. good catch

fsoubelet: Because it defaults to `now` anyway, right? I'll let you remove.

omc3/knob_extractor.py
fsoubelet: See 5 comments below before answering this 😬

omc3/knob_extractor.py
JoschD: It's only when LoggingDB is called. I checked. So import of pytimber should not interrupt workflow. I think what you are describing was a problem in earlier versions and might still be on in pjLSA, but not sure. Moved it up now to underline that it's an import.

fsoubelet (approved):

Mael-Le-Garrec (approved):

PR #398, opened 2022-09-22
cleaned logging

awegsche:
This cleans up the logging of knob extractor by suppressing pytimber logs and printing only one statement for missing knob values.

### DIFF
--- a/omc3/knob_extractor.py
+++ b/omc3/knob_extractor.py
@@ -67,6 +67,7 @@ from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path
 from typing import Dict, List, Optional, Sequence, Tuple, Union
+import logging
 
 import pandas as pd
 import tfs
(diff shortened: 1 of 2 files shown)


omc3/knob_extractor.py
```diff
@@ -279,16 +281,23 @@ def _extract(ldb, knobs_dict: KnobsDict, knob_categories: Sequence[str], time: d
...
+                LOGGER.debug(f"Value for {knob} is not a number or infinite")
+                knobs_nan_or_inf.append(knob)
                 continue
 
             LOGGER.info(f"Knob value for {knob} extracted: {value} (unscaled)")
             knobs[knob].value = value
+
+    if len(knobs_nan_or_inf):
+        LOGGER.info(f"The following knobs didn't return a value (or NaN/Inf):")
+        LOGGER.info(f" {', '.join(knobs_nan_or_inf)}")
```
fsoubelet, line 299: This does not work well with our logging setup as we append text at the right of this.

awegsche: you mean the `:`?
then just delete it? I thought it would be nice to split it up for readability or even log each missing knob in it's own log line

fsoubelet: I mean it will end up looking like this, iirc:
```
   INFO | The following knobs didn't return a value (or NaN/Inf): | .knob_extractor
   INFO | knob1, knob2, knob3, etc | .knob_extractor
```

awegsche: that's how it would look like, yes

awegsche:
removed the whole nan/inf/missing knobs logging. just showing the one that _have_ a value is probably enough

fsoubelet (approved):

JoschD (approved):

PR #400, opened 2022-09-30
Added IP2 and IP8 + statetracker State

JoschD:
If someone could write some tests for the State-Thing in the manner of the existing ones, including one that can run in the CERN network,  that would be great.

Log Output looks like:

```
   INFO | ---- STATE @ 2022-09-30 14:54:13.802200 ---- | .knob_extractor
   INFO | Optics:       R2022a_A11mC11mA10mL10m | .knob_extractor
   INFO | Beam Process: RAMP-SQUEEZE-6.8TeV-ATS-1.3m_V1@0_[START] | .knob_extractor
   INFO | Optics ID:    6912.0 | .knob_extractor
   INFO | HyperCycle:   6.8TeV_2022 | .knob_extractor
```


Mael-Le-Garrec (dismissed):

tests/unit/test_knob_extractor.py
```diff
@@ -437,6 +443,24 @@ def test_extractor_in_cern_network(self, tmp_path, knob_definitions, saved_knobf
...
+        state_df = main(time=time, state=True)
+
+        # format a bit to make frames equal
+        state_df = state_df.applymap(str)
+        state_df.headers[Head.time] = str(state_df.headers[Head.time])
+        state_df.index.name = "NAME"
+
+        # check frames
+        assert_dict_equal(state_df.headers, old_df.headers)
+        assert_frame_equal(state_df, old_df)  # dframe_y already pandas
```
fsoubelet, line 462: That's not really gonna work everytime, is it? What if we're not on 30cm optics? Best we can do I guess?

fsoubelet, line 462: can remove mention of `dframe_y`

fsoubelet (approved):

tests/unit/test_knob_extractor.py
JoschD: It is. it checks it at a given time-stamp, which is always the same.

PR #403, opened 2022-10-12
[Bugfix] fix phase filtering

awegsche:
If the phase matrix is filtered, both rows **and** columns have to be filtered.


JoschD (approved):

fsoubelet (dismissed):

tpersson (dismissed):
Excellent! and we have seen that this indeed solve the issue!

fsoubelet (approved):
Updated version and changelog so we can release asap. Will merge once tests pass since everyone already approved.

PR #404, opened 2022-11-01
Minor fixes in fake measurement, full response and corrections

JoschD:
Fullresponse Path was not converted to Path() instance. Fixed


fsoubelet (approved):

Mael-Le-Garrec (approved):
I approve of those present changes.

Issue #407, opened 2022-11-10
[Documentation]: Clean up after updates


JoschD:
### Documentation

- [X] Yes

### Pertinent Section

https://pylhc.github.io/omc3/

### Suggested Changes or Additions

- [ ] Fix the Navbar links (they are now two layers instead of one)
- [ ] Fix the expansion of `ArrayLike` typehints
- [ ] Check all math-formulas for proper tex-ing
- [ ] Add reference to Josch's b6 paper (for the ampdet/bbq analysis)
- [ ] Add local Table Of Contents via css: https://github.com/pylhc/irnl_rdt_correction/blob/restructuring/doc/_static/css/custom.css

fsoubelet:
The second one was solved in #413

JoschD:
5th point fixed in #421

PR #413, opened 2023-04-20
Check Corrections 

JoschD:
Implements the feature of checking corrections and their influence on the optics via MAD-X and plotting the expected results.


JoschD:
I apologize for the length of this PR... 
There is no version bump yet, as I want to do the release later, when the GUI itself is ready.

omc3/correction/utils_check.py
```diff
@@ -0,0 +1,34 @@
...
+    params.add_parameter(name="individual_to_input",
+                         action="store_true",
+                         help="Save plots for the individual corrections "
+                              "into the corrections input folders. "
+                              "Otherwise they go with suffix into the output_folders."
+                         )
+    params.update(get_optics_style_params())
+    params.update(get_plottfs_style_params())
+    params["plot_styles"]["default"] = params["plot_styles"]["default"] + ["correction_test"]
+    return params
```
fsoubelet, line 34: I would rather have this file be `omc3.utils.correction`

omc3/global_correction.py
```diff
@@ -47,7 +47,7 @@
 
 *--Optional--*
 
-- **beta_file_name**:
+- **beta_filename**:
```
fsoubelet, line 50: Is this gonna break something from the side of the GUI calling?

omc3/model/accelerators/lhc.py
```diff
@@ -200,21 +200,39 @@ def beam(self, value) -> None:
     def get_lhc_error_dir() -> Path:
         return LHC_DIR / "systematic_errors"
 
-    def get_variables(self, frm=None, to=None, classes=None):
+    def get_variables(self, frm: float = None, to: float = None, classes: Iterable[str] = None):
```
fsoubelet, line 203: I would personnally use `from_` but that's just a suggestion

omc3/utils/iotools.py
```diff
@@ -132,7 +175,26 @@ def remove_none_dict_entries(dict_: dict) -> dict:
     return {key: value for key, value in dict_.items() if value is not None}
 
 
-def save_config(output_dir: Path, opt: dict, script: str):
+def maybe_add_command(opt: dict, script: str) -> dict:
```
fsoubelet, line 178: Did not understand 😬

tests/accuracy/test_global_correction.py
```diff
@@ -64,62 +86,68 @@ def test_lhc_global_correct(tmp_path, model_inj_beams, orientation):
     """Creates a fake measurement from a modfied model-twiss with (skew)
     quadrupole errors and runs global correction on this measurement.
     It is asserted that the resulting model approaches the modified twiss.
+    In principle one could check one further model, build from the final
```
fsoubelet, line 89: "one could check one"?

fsoubelet:
Unrelated but since I saw some `ArrayLike` instances in there and it relates to #407, I have found the fix for their look in the documentation.

It is solved by adding the following in our `doc/conf.py`:
```python
# This is to tell Sphinx how to print some specific type annotations
# See: https://stackoverflow.com/a/67483317
# See: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases
autodoc_type_aliases = {"ArrayLike": "ArrayLike"}
```

This way they will appear as literally `ArrayLike` in the documentation.

omc3/global_correction.py
JoschD: yes, but I changed the `max_iter` to `iterations` as well, so I have updated that already in the branch I am working on

omc3/correction/utils_check.py
JoschD: I disagree, as the logic at the moment is: `omc3.utils` -> standalone utils. Everything else is mostly contained within the module itself. And as it is only used there, I think it should live in this module. BUT I am not very happy with the naming. Suggestions welcome.

omc3/correction/filters.py
```diff
@@ -90,17 +91,25 @@ def _get_filtered_generic(col: str, meas: pd.DataFrame, model: pd.DataFrame, opt
...
     # if opt.automatic_model_cut:  # TODO automated model cut
     #     model_filter = _get_smallest_data_mask(np.abs(meas.loc[:, f"{DELTA}{col}"].to_numpy()), portion=0.95)
     if f"{PHASE}" in col:
         new[NAME2] = meas.loc[:, NAME2].to_numpy()
-        second_bpm_in = np.in1d(new.loc[:, NAME2].to_numpy(), new.index.to_numpy())
-        good_bpms = error_filter & model_filter & second_bpm_in
-        good_bpms[-1] = False
+        second_bpm_exists = np.in1d(new.loc[:, NAME2].to_numpy(), new.index.to_numpy())
+        good_bpms = error_mask & model_mask & second_bpm_exists
+        good_bpms[-1] = False  # TODO not sure why, ask Lukas? (jdilly)
```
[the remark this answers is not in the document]
JoschD: No, I only noticed yesterday. I was hoping I would come up with an explanation myself. I will ask around

omc3/check_corrections.py
```diff
@@ -0,0 +1,615 @@
...
+
+@entrypoint(get_params())
+def correction_test_entrypoint(opt: DotDict, accel_opt) -> None:
+    """ Test the given corrections.
+    TODO: Instead of writing everything out, it could return
+          dictionaries of the TFSDataFrames and Figures.
+          But I don't see a usecase at the moment (jdilly 2023)
+    """
+    LOG.info("Starting Correction Test.")
+    # save_config(Path(opt.output_dir), opt, __file__, unknown_opt=accel_opt)
```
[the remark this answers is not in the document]
JoschD: Ohh, thanks for noticing. I commented it out while testing some things. needs to be put back!

omc3/utils/iotools.py
```diff
@@ -79,18 +93,47 @@ def __new__(cls, value):
 class UnionPathStr(metaclass=get_instance_faker_meta(Path, str)):
     """A class that can be used as Path and string parser input, but does not convert."""
     def __new__(cls, value):
-        if isinstance(value, str):
-            value = value.strip("\'\"")  # behavior like dict-parser, IMPORTANT FOR EVERY STRING-FAKER
-        return value
+        return
```
JoschD, line 96: doesn't return anything?

omc3/utils/iotools.py
```diff
@@ -54,22 +59,31 @@ def copy_item(src_item, dest):
...
 
 class PathOrStrOrDataFrame(metaclass=get_instance_faker_meta(TfsDataFrame, Path, str)):
     """A class that behaves like a Path when possible, otherwise like a string."""
     def __new__(cls, value):
-        if isinstance(value, str):
-            value = value.strip("\'\"")  # behavior like dict-parser, IMPORTANT FOR EVERY STRING-FAKER
+        value = strip_quotes(value)
         try:
             return Path(value)
         except TypeError:
```
JoschD, line 89: should return the dataframe!

omc3/utils/iotools.py
JoschD: Ah! This is actually a feature that came after a comment from you. And I am very proud of it.
This adds a new entry to the opt-dict called ";command" which contains the command used to run the current python script (from sys.argv). So this is similar to what was written in the `command.run` and you should be able to run this command and reproduce the current run/output. When the opt-dict is written out to an `.ini` the ";" at the beginning will cause this to be a comment in the ini, so you don't have to worry about issues when loading that file. But now you have the full run command in every saved ini file :)

JoschD: the "only if the executed file equals the given script." part means, that if you run your own script (`run_many_global_corrections.py`), which e.g. calls `global_correction` with different parameters, the command it not added in the case where `global_correction` calls this function, as the commandline parameters are probably completely different from the values written in the ini file.
So .. you can think of it a bit like `if __name__ == "__main__"` but a few levels deep.

fsoubelet: Very nice! I would suggest adding aaaaaaall of that text into the function docstring so it stays with the code ;)

JoschD: I KNEW you would say that xD

fsoubelet: You've already written it anyway, now it's mostly a copy paste :D

omc3/correction/filters.py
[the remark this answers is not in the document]
JoschD: Actually, why not doing it here? ^^ 
Hey @lmalina do you remember, why for the phase-advance the last entry is always removed when calculating the corrections? Is it unreliable?

omc3/correction/utils_check.py
fsoubelet: Fair enough.

omc3/model/accelerators/lhc.py
JoschD: leaving it for now as this is not new in this PR and I don't want to deal with any consequences (but actually I don't think we are using it)

alejandroborj (dismissed):

omc3/check_corrections.py
```diff
@@ -0,0 +1,612 @@
+"""
+Correction Test
+---------------
+
+Run a test, i.e. a MAD-X simulation to check how well the correction settings work.
```
fsoubelet, line 5: Since this text is for the documentation, I feel like it wouldn't hurt to give a little description of how the process works here.

omc3/check_corrections.py
```diff
@@ -0,0 +1,612 @@
...
+    return params
+
+
+@entrypoint(get_params())
+def correction_test_entrypoint(opt: DotDict, accel_opt) -> None:
+    """ Entrypoint function to test the given corrections.
+
+    .. todo:: Instead of writing everything out, it could return
+             dictionaries of the TFSDataFrames and Figures.
+             But I don't see a usecase at the moment (jdilly 2023)
```
fsoubelet, line 283: Could this be an optional parameter then? Defaults to False, but should someone want to use the function in another script they can.

omc3/check_corrections.py
```diff
@@ -0,0 +1,612 @@
...
+            meas_masks,
+        )
+
+    if opt.plot:
+        _do_plots(corrections, opt)
+
+
+# Input Parameters -------------------------------------------------------------
+
+def _check_opt_add_dicts(opt: DotDict) -> DotDict:  # acts inplace...
```
fsoubelet, line 322: Do a `copy.deepcopy` first to not act inplace? The function returns anyway, which really suggests it doesn't act inplace.

omc3/check_corrections.py
```diff
@@ -0,0 +1,612 @@
...
+def _glob_regex_paths(path: Path, pattern: str) -> List[Path]:
+    """ Filter the files in path by pattern and return a list of paths. """
+    return [path / f for f in glob_regex(path, pattern)]
+
+
+# Main and Output --------------------------------------------------------------
+
+def _get_measurement_filter(nominal_model: TfsDataFrame, opt: DotDict) -> Dict[str, pd.Index]:
+    """ Writes the filtered original data out, so that it can be easily plotted against
+    the corrections later. We now have the same BPMs and Column-Names. """
```
fsoubelet, line 384: If I understand correctly: this loads the measurements for different optics properties from the measurement folder, then applies a filter based on the params in `opt` and makes sure we have the same index (so all in all, the same common filtering) for all of them?

I would like to have this explanation in the docstring, even though it's a private function.

omc3/check_corrections.py
```diff
@@ -0,0 +1,612 @@
...
+    folder(s). """
+    if correction_name:
+        output_dir = output_dir / correction_name
+    LOG.info(f"Checking correction for {output_dir.name}")
+
+    # Created matched model
+    corr_model_path = output_dir / MODEL_MATCHED_FILENAME
+    corr_model_elements = global_correction._create_corrected_model(corr_model_path, correction_files, accel_inst)  # writes out twiss file!
+    corr_model_elements = _maybe_add_coupling_to_model(corr_model_elements, measurement)
+    LOG.debug(f"Matched model created in {str(corr_model_path)}.")
```
fsoubelet, line 427: Maybe this could be `corr_model_path.absolute()`?

omc3/check_corrections.py
```diff
@@ -0,0 +1,612 @@
...
+    }
+
+    LOG.debug(f"Filters selected for keys: {list(filter_dict.keys())}")
+    return filter_dict
+
+
+def _create_model_and_write_diff_to_measurements(
+        output_dir: Path, measurement: OpticsMeasurement, correction_name: str, correction_files: Sequence[Path],
+        accel_inst: Accelerator, rms_masks: Dict) -> OpticsMeasurement:
+    """ Create a new model with the corrections (well, the "matchings") inserted and calculate
```
fsoubelet, line 416: I would use `applied` instead of `inserted`

omc3/check_corrections.py
```diff
@@ -0,0 +1,612 @@
...
+    the difference to the measurements. This will be written out then into individual tfs-files in the output
```
fsoubelet, line 417: I would also specify that we calculate the difference to the nominal model and then apply this difference on the measurement to "predict" the effect of the correction on the measured machine (if I am correct about that), not that we calculate a difference from matched model to measurements as the sentence implies.

omc3/check_corrections.py
```diff
@@ -0,0 +1,612 @@
...
+                diff_models=diff_models,
+                colmap_meas=cols,
+                colmap_model=cols,
+                attribute=attribute,
+                rms_mask=rms_mask,
+            )
+    return output_measurement
+
+
+def _create_check_columns(measurement: OpticsMeasurement, output_measurement: OpticsMeasurement, diff_models: TfsDataFrame,
```
fsoubelet, line 484: Type hint that we do not return (`-> None`)

fsoubelet: Nice function though!

fsoubelet (commented):
I know I'm annoying, but you love it.

omc3/check_corrections.py
JoschD: No, no additional parameter. Just return it always. Doesn't cost anything, but I was to lazy to implement it. Can be done by whoever needs it then

omc3/check_corrections.py
JoschD: I think I accidentally copied this docstring from somewhere. This docstring makes no sense for this function

JoschD: (or maybe it was an old version of this function)

omc3/check_corrections.py
JoschD: I don't trust deepcopies. But I did as you asked

fsoubelet (approved):

alejandroborj (approved):

alejandroborj (approved):

PR #416, opened 2023-06-01
Plotting Windows

JoschD:
Adds QT-Based plotting windows, that allow sorting the opened plots better.

### DIFF
--- a/doc/modules/plotting.rst
+++ b/doc/modules/plotting.rst
@@ -56,3 +56,4 @@ Common Plot Functionality
 
 .. automodule:: omc3.plotting.utils.windows
     :members:
+    :noindex:
--- a/tests/unit/test_check_corrections.py
+++ b/tests/unit/test_check_corrections.py
@@ -301,10 +301,10 @@ class TestPlotting:
                     rdt = "f1001" if idx_tab == 1 else "f1010"
                     assert len(title_parts) == 2
                     assert title_parts[0] == rdt
-                    assert title_parts[1].lower() == ["ap", "ri"][idx]
+                    assert title_parts[1].lower() == ["a/p", "r/i"][idx]
             else:
                 for idx, plottab in enumerate(tab.tabs):
-                    assert plottab.title.lower() == ("f1010 ri" if idx else "f1001 ap")
+                    assert plottab.title.lower() == ("f1010 r/i" if idx else "f1001 a/p")
 
 
     @pytest.mark.basic


setup.py
```diff
@@ -44,6 +44,8 @@
     ],
     "optional": [
         "pymupdf>=1.22.0",  # logbook for conversion from pdf to png
+        "qtpy>=2.3.1",  # putting plots in windows
+        "pyside2>=5.15.0.0",  # putting plots in windows
```
fsoubelet, line 48: Why do we force this? I thought the point of `qtpy` was that it would pick up any bindings on the local machine and make do with that for us. Shouldn't we just have `qtpy` as a dependency?

fsoubelet: Also, `pyside2` is under `LGPL` and I'm still not convinced the optional dependency trick really holds, given we're `MIT` licensed and want to stay that way.

fsoubelet: Also (yeah I know), `pyside2` provides bindings to `Qt5` which [has already reached end of life](https://endoflife.date/qt). So if we're forcing bindings, shouldn't we go for `pyside6` then?

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
+"""
+Plot Window
+------------
+
+In this module different classes are defined, allowing to put plots
```
fsoubelet, line 5: Let's specify this work for `matplotlib` figures, not simply "plots"

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
+"""
+Plot Window
+------------
+
+In this module different classes are defined, allowing to put plots 
+manually into windows.
```
fsoubelet, line 6: GUI windows?

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+LOG = logging_tools.get_logger(__name__)
+
+# --- optional qtpy import block -----------------------------------------------
+
+try:
+    from qtpy.QtWidgets import (
+        QApplication, QMainWindow, QTabWidget, QVBoxLayout, QWidget
+    )
+except ImportError as e:
+    LOG.debug(f"Could not import QtPy: {str(e)}")
```
fsoubelet, line 38: Let's make this a warning maybe?

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+    QMainWindow, QApplication, QVBoxLayout  = None, None, None
+    QWidget, QTabWidget = object, object
+else:
+    from matplotlib.backends.backend_qt5agg import FigureCanvasQTAgg as FigureCanvas
+    from matplotlib.backends.backend_qt5agg import NavigationToolbar2QT as NavigationToolbar
```
fsoubelet, line 43: Again, Qt5 has reached EOL. The [matplotlib docs](https://matplotlib.org/stable/api/backend_qt_api.html) say that these modules are deprecated and here for backwards-comatibility. I think the recommended way is to use `matplotlib.backends.backend_qtagg` instead and let matplotlib figure out the Qt version and bindings itself.

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+    try:
+        matplotlib.use('Qt5agg')
```
fsoubelet, line 45: `Qtagg` I guess

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+class PlotWidget(QWidget):
+
+    def __init__(self, *figures: Figure, title: str):
+        """A derived QWidget that contains Matplotlib-Figures,
+        stacked vertically, each with its own toolbar.
+
+        Args:
+            *figures (Figure): Figures to be contained in the widget
+            title (str): Name of the widget
+            skip_super (bool): If True the superclass __init__ will be skipped (ONLY FOR TESTS)
```
fsoubelet, line 81: This doesn't seem to be an argument anymore, should be removed?

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        QWidget.__init__(self)   # no super(), because will not work with tests
+
+        self.title: str = title
+        self.figures: Tuple[Figure] = figures
```
fsoubelet, line 86: `Tuple[Figure, ...]`

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+        self._canvas_toolbar_handles: List[Tuple[FigureCanvas, NavigationToolbar]] = []
+
+        self._layout = QVBoxLayout()
+        self.setLayout(self._layout)
+
+        for figure in figures:
```
fsoubelet, line 92: in `self.figures` maybe? Otherwise I'n not sure why we cache then inside the object attributes

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+class TabWidget(QTabWidget):
+
+    def __init__(self, title: str = None):
+        """A simple tab widget, that in addition
+        to QTabWidget keeps track of the tabs via dictionary
+        and can have a title.
+        
+        Args:
+            title (str): Title of the created Window
+            skip_super (bool): If True the superclass __init__ will be skipped (ONLY FOR TESTS)
```
fsoubelet, line 110: Same here, absent in args

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+    def add_tab(self, widget: QWidget):
+        """ Add a new tab made from the given widget.
+
+        Args:
+            widget (QWidget): Widget to add.
+        """
+        self.tabs_widget.add_tab(widget)
+
+    def show(self):
+        self.app.exec_()
```
fsoubelet, line 156: Also related to Qt5 being EOL, if we want to go for Qt6 this would be `.exec()` then.

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+        self.title: str = title
+        self.tabs: Dict[str, QWidget] = {}
+
+    def add_tab(self, widget: QWidget):
+        self.tabs[widget.title] = widget
+        self.addTab(widget, widget.title)
+
+
+class SimpleTabWindow:
+
```
fsoubelet, line 123: This is a full-blown windows application that runs its event loop, right? Maybe let's give it a short docstring to specify that. It would help for Qt-noobs like me

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+            widget (QWidget): Widget to add.
+        """
+        self.tabs_widget.add_tab(widget)
+
+    def show(self):
+        self.app.exec_()
+
+
+class VerticalTabWindow(SimpleTabWindow):
+
```
fsoubelet, line 160: I feel like we should use some ascii-art for visualisation in the docs now 😅

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+        Args:
+            widget (QWidget): Widget to add.
+            tab (str): Title/Name of the tab .
+
+        """
+        self.tabs_widget.tabs[tab].add_tab(widget)
+
+    def show(self):
+        self.app.exec_()
```
fsoubelet, line 186: Idem than above

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+        self.tabs_widget.setTabPosition(QTabWidget.West)
+
+    def add_to_tab(self, widget: QWidget, tab: str):
+        """ Add `widget` to the tab with the given
+        `tab` identifier.
+
+        Args:
+            widget (QWidget): Widget to add.
+            tab (str): Title/Name of the tab .
+
```
fsoubelet, line 181: empty line

omc3/plotting/utils/windows.py
```diff
@@ -0,0 +1,208 @@
...
+        self.tabs_widget.tabs[tab].add_tab(widget)
+
+    def show(self):
+        self.app.exec_()
+
+
+def create_pyplot_window_from_fig(fig: Figure, title: str = None):
+    """Creates a window from the given figure, which is managed by pyplot. 
+    This is similar to how figures behave when created with `pyplot.figure()`,
+    but you can crate the figure instance first and the manager later.
```
fsoubelet, line 192: create

omc3/plotting/plot_tfs.py
```diff
@@ -531,8 +536,22 @@ def _create_plots(fig_collection, opt):
...
+            for fig_container in fig_collection.figs.values():
+                tab = PlotWidget(fig_container.fig, title=fig_container.title)
+                window.add_tab(tab)
+            window.show()
+            return
+        
+        if len(fig_collection) > rcParams['figure.max_open_warning']:
+            log_no_qtpy_many_windows()
+
+        for fig_container in fig_collection.figs.values():
```
fsoubelet, line 550: Is this needed if we `plt.show()` afterwards? As far as I understand this, `plt.show()` would spawn a window for each figure anyway? I guess what I don't understand here is the added value

omc3/plotting/plot_checked_corrections.py
```diff
@@ -152,29 +152,43 @@
...
+    FILE_COLUMN_MAPPING, ColumnsAndLabels, RDT_COLUMN_MAPPING, RDT_PHASE_COLUMN, 
+    RDT_IMAG_COLUMN, RDT_REAL_COLUMN, RDT_AMPLITUDE_COLUMN
+)
 from omc3.optics_measurements.constants import EXT
-from omc3.plotting.plot_optics_measurements import (_get_x_axis_column_and_label, _get_ip_positions,
-                                                    get_optics_style_params, get_plottfs_style_params)
+from omc3.plotting.plot_optics_measurements import (
+    _get_x_axis_column_and_label, _get_ip_positions,
+    get_optics_style_params, get_plottfs_style_params
+)
```
fsoubelet, line 176: And you hate when I format 😉

omc3/plotting/plot_checked_corrections.py
```diff
@@ -403,6 +429,72 @@ def save_plots(output_dir, figure_dict, input_dir=None):
             fig.savefig(output_path)
 
 
+def show_plots(figure_dict: Dict[str, Figure]):
+    """ Shows plots.
```
fsoubelet, line 433: Displays the provided figures.

omc3/plotting/plot_checked_corrections.py
```diff
@@ -403,6 +429,72 @@ def save_plots(output_dir, figure_dict, input_dir=None):
...
+    If qtpy is installed, they are shown in a single window.
```
fsoubelet, line 434: escape `qtpy` in ``?

omc3/plotting/plot_checked_corrections.py
```diff
@@ -403,6 +429,72 @@ def save_plots(output_dir, figure_dict, input_dir=None):
...
+    The individual corrections are sorted in to vertical tabs,
```
fsoubelet, line 435: into

omc3/plotting/plot_checked_corrections.py
```diff
@@ -403,6 +429,72 @@ def save_plots(output_dir, figure_dict, input_dir=None):
...
+        plt.show()
+        return
+
+    rdt_pattern = re.compile(r"f\d{4}")
+    rdt_complement = {
+        RDT_REAL_COLUMN.text_label: RDT_IMAG_COLUMN,
+        RDT_AMPLITUDE_COLUMN.text_label: RDT_PHASE_COLUMN,
+    }
+
+    correction_names = sorted(set([n.split(SPLIT_ID)[0] for n in figure_dict.keys() if SPLIT_ID in n]))
```
fsoubelet, line 455: Here n is the name, right? Let's rename accordingly

omc3/plotting/plot_checked_corrections.py
```diff
@@ -403,6 +429,72 @@ def save_plots(output_dir, figure_dict, input_dir=None):
...
+    for correction_name in [None] + list(correction_names):
+        if not correction_name:
+            parameter_names = iter(sorted(n for n in figure_dict.keys() if SPLIT_ID not in n))
```
fsoubelet, line 458: same

omc3/plotting/plot_checked_corrections.py
```diff
@@ -403,6 +429,72 @@ def save_plots(output_dir, figure_dict, input_dir=None):
...
+            correction_tab_name = "All Corrections"
+        else:
+            parameter_names = iter(sorted(n for n in figure_dict.keys() if correction_name in n))
```
fsoubelet, line 461: same

omc3/plotting/plot_checked_corrections.py
```diff
@@ -403,6 +429,72 @@ def save_plots(output_dir, figure_dict, input_dir=None):
...
+            correction_tab_name = correction_name
+
+        current_tab = TabWidget(title=correction_tab_name)
+        window.add_tab(current_tab)
+
+        for name_x in parameter_names:
+            tab_prename = name_x.split(SPLIT_ID)[-1]
```
fsoubelet, line 468: maybe add a comment to explain which part you're getting, not clear to me by just reading the code

omc3/plotting/plot_checked_corrections.py
```diff
@@ -403,6 +429,72 @@ def save_plots(output_dir, figure_dict, input_dir=None):
...
+
+            if rdt_pattern.match(tab_prename):
```
fsoubelet, line 470: Let's make a comment for this block. 

If I understand correctly we're getting RDT corrections, we split the RDT components (amplitude, phase, real, imag as far as I can tell?) and make a tab for each, which shows the various attempted corrections?

omc3/plotting/plot_checked_corrections.py
```diff
@@ -403,6 +429,72 @@ def save_plots(output_dir, figure_dict, input_dir=None):
...
+                else:
+                    name_y = "_".join(name_x.split("_")[:-1] + [complement_column.text_label,])
+
+                tab_name = f"{rdt} {column[0].upper()}{complement_column.text_label[0].upper()}"
+
+            else:
+                tab_name = " ".join(tab_prename.split("_")[:-1 if correction_name else -2])
+                name_y = next(parameter_names)
+
+            new_tab = PlotWidget(
```
fsoubelet, line 489: I'm confused why `new_tab` is not a `TabWidget`?

omc3/plotting/optics_measurements/utils.py
```diff
@@ -33,8 +34,9 @@ class DataSet:
 class FigureContainer:
     """Container for attaching additional information to one figure."""
     def __init__(self, id_: str, path: Path, axes_ids: Iterable[str]) -> None:
```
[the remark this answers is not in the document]
JoschD: I changed the field to `.id` as it is acutally only the window title if you want it to be. Not sure exactly what the difference between an identifyer and a title is, but this was meant as the former but can be abused as the latter...

omc3/plotting/utils/windows.py
JoschD: the docstring is in the init. but I added a sentence

omc3/plotting/utils/windows.py
JoschD: (This function) creates

omc3/plotting/plot_checked_corrections.py
JoschD: n is more of an id than a name

omc3/plotting/plot_checked_corrections.py
JoschD: because the TabWidget is a container for tabs. But each tab itself is just a widget.

omc3/plotting/plot_checked_corrections.py
JoschD: Well, that depends... because of the way plot-tfs creates unique IDs and what we are plotting.
There is two kinds of plots:
a) Plots showing the expected value after correction + the measured value. In this cases the unique id is created by the file-name the value comes from.
b) Plots showing expectation, measurement and correction for a single Correction. In this case all of these come from the same file, so the unique name is filename_columnname

omc3/plotting/optics_measurements/utils.py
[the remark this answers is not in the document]
fsoubelet: Maybe we can rename it "identifier" or something? I don't know why but a trailing underscore in a variable name is rough on my eyes

omc3/plotting/utils/windows.py
fsoubelet: Fair enough. I see you changed to use `self.figures` in the end, all fine by me here.

omc3/plotting/plot_checked_corrections.py
fsoubelet: As I hold the belief that k is a far superior letter to n, I'm fine with the change

setup.py
fsoubelet: Alright for LGPL. Regarding the following:
>I would be very annoyed if I would try to run the code with qtpy and it would still complain about missing dependencies that are not mentioned anywhere

I feel like going through `qtpy` means letting it handle the bindings and leaving the responsibility of having them installed to the user, while also being very clear about this in the docs.

As it's optional though it's not a huge deal for sure. But then having `pyside6` (cf [d5c0fb8](https://github.com/pylhc/omc3/pull/416/commits/d5c0fb8504207f2b2c4fb2d4f9ef0436e4ae4081)) specified feels like we're "forcing" (big airquotes here) some bindings on the user.

fsoubelet: To be clear, I'm not a fan but I don't consider this blocking.

omc3/plotting/plot_checked_corrections.py
JoschD: I actually explained it the wrong way around... but yeah... it's the ID and I am trying to extract a useful tab-name from there

fsoubelet (approved):
Works fine with `PyQt6` and `PySide6`, good with me 👍

PR #423, opened 2023-07-21
Added raw strings and/or backslashes

JoschD:
Fixes the deprecation warnings, which come up due to non-escaped backslashes in strings


fsoubelet (approved):
Thanks

fsoubelet:
@JoschD seems all good now with the `tfs` patch, I think you can merge this and then https://github.com/pylhc/omc3/pull/426

PR #425, opened 2023-08-28
Adjust correction-knob json files for 2023

JoschD:
- removed Q4Q5Q6 from default categories
- added categories with original knobs 
- added category for HB


awegsche (dismissed):
Looks good

fsoubelet (approved):
Good for me. 

On a side note, doesn't feel too intuitive that we're storing these in the 2012 folder, maybe we should have an admonition on the OMC website about this.

awegsche (approved):

PR #426, opened 2023-09-01
Plot normalized dispersion

JoschD:
Fixing the plotting of normalized dispersion optics measurements and adding a "labels' feature.


omc3/plotting/plot_optics_measurements.py
```diff
@@ -416,6 +514,45 @@ def _plot_param(optics_parameter, files, file_labels, x_column, x_label, ip_posi
                            'share_xaxis'])
     )
 
+def _plot_norm_dispersion(optics_parameter, files, file_labels, x_column, x_label, ip_positions, opt):
+    """Plotting normalized dispersion function for normalized dispersion.
+    Normalized dispersion is special, as we only have X. 
+    We therefore plot the delta in the second plot. """
```
fsoubelet, line 520: Bit repetitive with "normalized dispersion". For the second plot, do you mean second axis?

fsoubelet (dismissed):

omc3/plotting/plot_optics_measurements.py
awegsche:
```python
 """Plotting normalized dispersion function for normalized dispersion.
    Normalized dispersion is special, as we only have horizontal normalized dispersion. 
    We therefore plot the delta of normalized dispersion in the second plot in normalized dispersion plots. """
    ```

awegsche: With second plot, Josch probably means the bottom plot of our standard two-plot windows in the GUI (and two-plot matplotlib outputs), usually showing X / Y, real / imag, or amp / phase.

+1 for using this unused space for the delta.

fsoubelet (approved):

PR #427, opened 2023-09-06
add get_bad_bpms

awegsche:
# New Feature

This script collects a list of all BPMs declared as bad by either SVD cleaning or isolation forest and compiles a list with number of occurances for each BPM that has been declared at least 50% (25%) of the time.

# Purpose

Every now and then BI asks for a list of BPMs that cause trouble for our measurements so that they can take a closer look during the next [(E?YET|L)S](https://regex101.com/r/ShN5hC/3) . This script is intended to facilitate the process of retrieving such a list.

# Notes

This script is just a template, if specific measurements / globbing sounds better than just taking unchecked all the measurements in one (or several) output folder, feel free to adapt if necessary

I am not going to write tests for this. It's a script that's run once every 2-3 years.


omc3/scripts/get_bad_bpms.py
```diff
@@ -0,0 +1,150 @@
...
+        return None
+
+    measurements_svd = [get_bad_bpms(p) for p in measurements]
+    measurements_svd = [p for p in measurements_svd if p is not None]
+
+    measurements_iforest = [get_iforest_bpms(p) for p in measurements]
+    measurements_iforest = [p for p in measurements_iforest if p is not None]
+    return (measurements_svd, measurements_iforest)
+
+def get_bad_bpms_for_beam_and_plane_svd(date: str, plane: str, beam: str, bad_bpms_list):
```
fsoubelet, line 69: Feels like `bad_bpms_list` is not expecting a list

omc3/scripts/get_bad_bpms.py
```diff
@@ -0,0 +1,150 @@
...
+            if "model" in words and "not" in words:
+                continue
+            appeared.append(words[0])
+            if words[0] in bad_bpms_list:
+                bad_bpms_list[words[0]] = bad_bpms_list[words[0]] + 1
+            else:
+                bad_bpms_list[words[0]] = 1
+    return len(measurements)
+
+def get_bad_bpms_for_beam_and_plane_iforest(date: str, plane: str, beam: str, bad_bpms_list):
```
fsoubelet, line 87: Same

fsoubelet (commented):
Details here and there. I'm open to contributing about them if you would like.

I'm wondering if this should be included in `omc3` itself. Maybe this is a good fit for `pylhc`? Would like @JoschD's opinion.

Also, in hindsight to my review in [#pylhc/py](https://github.com/pylhc/pylhc.github.io/pull/100), I like the little `purpose` paragraph you have in this PR. Maybe it could be copy-pasted or adapted there?

JoschD:
Please don't make me write tests :-/

JoschD (changes requested):
Docstring for the args missing

omc3/scripts/bad_bpms_summary.py
```diff
@@ -0,0 +1,294 @@
...
+        name="accel_glob",
+        type=str,
+        default="LHCB*",
+        help="Accelerator name (glob for the sub-directories)."
+    )
+    return params
+
+
+@entrypoint(get_params(), strict=True)
+def bad_bpms_summary(opt: DotDict):
```
jgray-19, line 79: Is there any point in doing anything if opt.outfile and opt.print_percentage are both none? Could emit a warning or error?

omc3/scripts/bad_bpms_summary.py
```diff
@@ -0,0 +1,294 @@
+"""
+Bad BPMs Summary
+----------------
+
+Scans all measurements in a list of given GUI output folders and compiles a list of bad BPMs with
+their given number of appearances after 'harpy' and 'isolation forest'.
+"""
+from collections.abc import Sequence
```
jgray-19, line 8: No `from __future__ import annotations`?

omc3/scripts/bad_bpms_summary.py
```diff
@@ -0,0 +1,294 @@
...
+    # Count how often a BPM is bad
+    df_counted = df.groupby([NAME, ACCEL, SOURCE, PLANE]).size().reset_index(name=COUNT)
+
+    # Count the total number of (unique) files for each combination of accelerator, source and plane
+    file_count = df.groupby([ACCEL, SOURCE, PLANE])[FILE].nunique().reset_index(name=FILE_COUNT)
+    df_counted = df_counted.merge(file_count, how="left", on=[ACCEL, SOURCE, PLANE])
+
+    df_counted.loc[:, PERCENTAGE] = round(
+        (df_counted[COUNT] / df_counted[FILE_COUNT]) * 100, 2
+    )
```
fsoubelet, line 264: Very nice

fsoubelet (changes requested):
Cool, mostly would like to see the module docstring include script parameters and potentially an example.

Worth noting: with the amount of `tfs.concat` we're doing here, without providing any setup for the headers handling, the results will always be headers-less. Not sure if it's something we care, or it we want to write some basic info in the final output's headers.

omc3/scripts/bad_bpms_summary.py
```diff
@@ -0,0 +1,294 @@
...
+    Args:
+        date_dir (Path): Path to the date directory.
+        accel_glob (str): Accelerator name (glob for the sub-directories).
+
+    Returns:
+        tfs.TfsDataFrame: TfsDataFrame with all bad-bpms for the date.
+    """
+    dfs: list[tfs.TfsDataFrame] = []
+
+    for accel_dir in date_dir.glob(accel_glob):
```
jgray-19, line 140: This loop seems specific to LHC? Do other machines have the same path like `root/date/accel/measurement`?

omc3/scripts/bad_bpms_summary.py
```diff
@@ -0,0 +1,294 @@
...
+    """ Reads a harpy bad-bpm file and returns a TfsDataFrame with all unique bad-bpms.
+    
+    Args:
+        svd_file (Path): Path to the bad-bpm file.
+
+    Returns:
+        tfs.TfsDataFrame: TfsDataFrame with all unique bad-bpms.
+
+    """
+    TO_INGNORE = ["not found in model", "known bad bpm"]
```
jgray-19, line 198: Ignore?

omc3/scripts/bad_bpms_summary.py
```diff
@@ -0,0 +1,294 @@
...
+    Returns:
+        tfs.TfsDataFrame: TfsDataFrame with all bad-bpms within selcted dates.
+    
+    """
+    dfs = []
+    for date in dates:
+        date_dir = root / date
+        if date_dir.is_dir():
+            dfs.append(collect_date(date_dir, accel_glob))
+            continue
```
jgray-19, line 121: I would prefer an `else` here, but up to you

omc3/scripts/bad_bpms_summary.py
```diff
@@ -0,0 +1,294 @@
...
+    for accel in sorted(df_counted[ACCEL].unique()):
+        accel_mask = df_counted[ACCEL] == accel
+        for source in sorted(df_counted[SOURCE].unique()):
+            source_mask = df_counted[SOURCE] == source
+            df_filtered = df_counted.loc[percentage_mask & source_mask & accel_mask, :]
+            msg = "\n".join(
+                f"{row[NAME]:>20s} {row[PLANE]}: {row[PERCENTAGE]:5.1f}% ({row[COUNT]}/{row[FILE_COUNT]})" 
+                for _,row in df_filtered.iterrows()
+            )
+            LOG.info(f"Highest bad BPMs of {accel} from {source}:\n" + msg)
```
jgray-19, line 288: Does this print to console always? If not, is this what the user would expect from the doc?

omc3/scripts/bad_bpms_summary.py
```diff
@@ -0,0 +1,294 @@
...
+
+# Collection of Data ---
+
+def get_empty_df() -> tfs.TfsDataFrame:
+    """ Create an empty TfsDataFrame with the correct column names. """
+    return tfs.TfsDataFrame(columns=[NAME, ACCEL, PLANE, SOURCE, FILE])
+
+
+def collect_bad_bpms(root: Path, dates: Sequence[Path | str], accel_glob: str) -> tfs.TfsDataFrame:
+    """ Create a TfsDataFrame with all bad-bpms within selcted dates.
```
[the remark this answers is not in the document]
jgray-19: Also line 113

omc3/scripts/bad_bpms_summary.py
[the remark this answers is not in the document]
JoschD: but they are not bad because the analysis says they are bad. They are bad, because we say they are bad. It's a bit of a self-fullfilling prophecy then.... but I have an idea

omc3/scripts/bad_bpms_summary.py
JoschD: yes, this structure is specific to the GUI which we use for all accelerators

omc3/scripts/bad_bpms_summary.py
JoschD: The logging setup is done in `get_logger(__name__)`. So if you run this as a script, this is always printed. If you import the function and have no logger set up, then not. I will adapt

omc3/scripts/bad_bpms_summary.py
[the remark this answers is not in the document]
fsoubelet: True, I guess it's not obvious here. Let's see your idea.

omc3/scripts/bad_bpms_summary.py
```diff
@@ -0,0 +1,412 @@
...
+            for data_dir in analysis_stage_dir.iterdir():
+                if not data_dir.is_dir():
+                    continue
+
+                df_collected = collect_bad_bpm_files_in_dir(data_dir)
+                if df_collected is not None:
+                    df_collected.loc[:, ACCEL] = accel_dir.name
+                    dfs.append(df_collected)
+
+    if not len(dfs):
```
jgray-19, line 231: Correct me if I'm wrong, but would `not dfs` be better instead here?

JoschD: but you will find different flavours of this depending on who writes the code

fsoubelet (approved):

jgray-19 (dismissed):

JoschD (approved):

PR #428, opened 2023-09-20
Hl fixes

awegsche:
this adds a couple of fixes related to hl

- important phase advances are not yet defined for HL

"The names of tertiaries are different in HL and in general everything in between q5 to Q5 is different. The phase advances for HL are likely different from LHC. We will finalise likely towards oct-nov 2023 after we will perform further optimizations."
-- Riccardo De Maria

- the response creator used the formatting `+f` for `delta_k`, which only printed exactly 6 decimal spaces. consequently, values below 1e-6 were written as 0 into the job file.


awegsche:
how does this change break plotting?

omc3/model/accelerators/lhc.py
```diff
@@ -339,6 +339,10 @@ def get_exciter_bpm(self, plane: str, commonbpms: List[str]):
         return None
 
     def important_phase_advances(self) -> List[List[str]]:
+        if "hl" in self.year:
```
fsoubelet, line 342: To be sure, maybe let's search in `self.year.lower()`

omc3/model/accelerators/lhc.py
```diff
@@ -339,6 +339,10 @@ def get_exciter_bpm(self, plane: str, commonbpms: List[str]):
         return None
 
     def important_phase_advances(self) -> List[List[str]]:
+        if "hl" in self.year: 
+            # skip if HiLumi
```
fsoubelet, line 343: Would you add a TODO in this comment? Would make it easier to come back to when Riccardo and team are ready with this info

omc3/correction/response_madx.py
```diff
@@ -99,9 +99,9 @@ def _generate_madx_jobs(
                 break
             var = variables[var_idx]
             incr_dict[var] = delta_k
-            current_job += f"{var}={var}{delta_k:+f};\n"
+            current_job += f"{var} = {var}{delta_k:+.15f};\n"
             current_job += f"twiss, file='{str(temp_dir / f'twiss.{var}')}';\n"
-            current_job += f"{var}={var}{-delta_k:+f};\n\n"
+            current_job += f"{var} = {var}{-delta_k:+.15f};\n\n"
```
JoschD, line 104: why not a more flexible format like e?

fsoubelet: Actually if we want to be sure, we can use the `#` option which outputs digits untils there are not more.

fsoubelet: Still needs to decide on a number of decimals to show before the power of 10

awegsche: I propose `{:.15e}`. this should print enough decimals for the given precision.

or we change `delta_k` to a string. this sounds like the most practical solution in this case :p

JoschD (approved):

omc3/correction/response_madx.py
JoschD: I don't think anyone will ever need that much precision (i.e. there will be no difference in the machine) and probably 4 digits would already be extensive, but as I don't see any disadvantage either ... resolved

fsoubelet (approved):

PR #431, opened 2023-11-23
Extend Harpy naming capabilities

JoschD:
Adds a suffix-parameter to harpy, so that the user can change the output filename for the analysis files.
I was considering making it so, that you can specify the full filename... but I think adding only a suffix is okay for now.

### DIFF
--- a/omc3/hole_in_one.py
+++ b/omc3/hole_in_one.py
@@ -33,7 +33,7 @@ import os
 from collections import OrderedDict
 from copy import deepcopy
 from datetime import datetime
-from os.path import abspath, basename, dirname, join
+from os.path import abspath, basename, dirname, join, splitext
 
 import turn_by_turn as tbt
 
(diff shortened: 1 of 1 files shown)


omc3/hole_in_one.py
```diff
@@ -370,7 +378,7 @@ def _run_harpy(harpy_options):
         tbt_datas = [(tbt.read_tbt(option.files, datatype=option.tbt_datatype), option) for option in all_options]
         for tbt_data, option in tbt_datas:
             lins.extend([handler.run_per_bunch(bunch_data, bunch_options)
-                         for bunch_data, bunch_options in _multibunch(tbt_data, option)])
+                         for bunch_data, bunch_options in _add_suffix_and_loop_over_bunches(tbt_data, option)])
```
Mael-Le-Garrec, line 381: change `loop` for something like `iter` so we know you can iterate on it

Mael-Le-Garrec (changes requested):

fsoubelet (approved):

PR #434, opened 2024-02-13
Fix pandas indexing error

Mael-Le-Garrec:
The pandas Series has to be converted to a numpy array for the indexing to work.
Functionality has been removed after depreciation and leads to an error.
Somehow the change has already been made for the function `calculate_dpoverp` with the ~same code at line 76


fsoubelet:
Thanks! Could you change the version for a patch release @Mael-Le-Garrec ?

fsoubelet (approved):

PR #436, opened 2024-02-23
Fix CO units issue in BBsrc converter

### LINKED ISSUE #435: [Bug]: CO in BBS to OMC3 converter wrong units
### Documentation

- [X] Yes

### Operating System

all

### Python Version

all

### Package Version

0.3.0

### Bug Description

When using the bbs to omc3 converter, the closed orbit units are not converted. They are mm in bbs but m in omc3.

### Steps to Reproduce

_No response_

### Relevant output

_No response_

### Possible Fix Implementation

_No response_

fsoubelet:
Closes #435.

Fix is simply to divide by 1e3 when doing the conversion to go from the `mm` of BBsrc to the `m` of omc3.


JoschD (approved):
The multiplication looks very efficient. Much better than division

PR #440, opened 2024-06-05
Bugfix MQSORG knob naming in beam 2

JoschD:


fsoubelet:
@JoschD considering Python 3.8 is close to the end of its security support (a few months), maybe we could drop it from the test matrices? Seems starting with 3.9 tables does not cause issues installing on macOS.

fsoubelet (approved):

PR #441, opened 2024-07-08
Use modifiers for knobs madx

JoschD:
Instead of having the `knobs.madx` hard-coded they will be given as modifiers from the GUI.
This avoids having to create empty `knobs.madx` file beforehand, whenever you want to create a model.

This PR also cleans up a few more things from the model creation:
- remove the `corrections.madx` from best-knowledge model, as we never used this. 
  If needed in the future, this can also be simply given as a modifier.
- Zip the mad-x output log-file from the response model creation. As this runs in parallel by default, using the number of available cores, the final (merged) log file can be quite big, even with the sequence loading not echoed.
- Also response creation: keep the 0th job-file as reference in the model directory, so that one can check that the setup of the machine is correctly performed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
     - removed `corrections.madx` from `lhc` best-knowledge model
     - zip up log-output files in `response_madx.py`
     - keep 0th output file in `response_madx.py` for reference of the model setup
+    - Sequence and modifiers use the acc-models symlink in madx-jobs where applicable.
 
 #### 2024-06-05 - v0.14.1 - _jdilly_
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
     - removed hard-coded `knobs.madx` from `lhc`
     - removed `corrections.madx` from `lhc` best-knowledge model
     - zip up log-output files in `response_madx.py`
-    - keep 0th output file in `response_madx.py` for reference
+    - keep 0th output file in `response_madx.py` for reference of the model setup
 
 #### 2024-06-05 - v0.14.1 - _jdilly_
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # OMC3 Changelog
 
+#### 2024-07-08 - v0.15.0 - _jdilly_
+
+- PINNING NUMPY TO < 2.0.0
+
+- Changed:
+  - Model creation:
+    - removed hard-coded `knobs.madx` from `lhc`
+    - removed `corrections.madx` from `lhc` best-knowledge model
+    - zip up log-output files in `response_madx.py`
+    - keep 0th output file in `response_madx.py` for reference
+
 #### 2024-06-05 - v0.14.1 - _jdilly_
 
 - Fixed:
--- a/omc3/knob_extractor.py
+++ b/omc3/knob_extractor.py
@@ -459,16 +459,16 @@ def _write_knobsfile(output: Union[Path, str], collected_knobs: tfs.TfsDataFrame
 def _get_knobs_def_file(user_defined: Optional[Union[Path, str]] = None) -> Path:
     """ Check which knobs-definition file is appropriate to take. """
     if user_defined is not None:
-        LOGGER.info(f"Using user defined knobs defition file: '{user_defined}")
+        LOGGER.info(f"Using user knobs-definition file: '{user_defined}")
         return Path(user_defined)
 
     if KNOBS_FILE_ACC_MODELS.is_file():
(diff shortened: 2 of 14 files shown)


omc3/knob_extractor.py
```diff
@@ -461,16 +459,16 @@ def _write_knobsfile(output: Union[Path, str], collected_knobs: tfs.TfsDataFrame
 def _get_knobs_def_file(user_defined: Optional[Union[Path, str]] = None) -> Path:
     """ Check which knobs-definition file is appropriate to take. """
     if user_defined is not None:
-        LOGGER.info(f"Using user defined knobs.txt: '{user_defined}")
+        LOGGER.info(f"Using user defined knobs defition file: '{user_defined}")
```
fsoubelet, line 462: Typo

omc3/knob_extractor.py
```diff
@@ -461,16 +459,16 @@ def _write_knobsfile(output: Union[Path, str], collected_knobs: tfs.TfsDataFrame
...
 
     if KNOBS_FILE_ACC_MODELS.is_file():
-        LOGGER.info(f"Using model folder's knobs.txt: '{KNOBS_FILE_ACC_MODELS}")
+        LOGGER.info(f"Using model folder's knobs.txt as definition file: '{KNOBS_FILE_ACC_MODELS}")
         return KNOBS_FILE_ACC_MODELS
 
     if KNOBS_FILE_AFS.is_file():
         # if all fails, fall back to lhc acc-models
-        LOGGER.info(f"Using fallback knobs.txt: '{KNOBS_FILE_AFS}'")
+        LOGGER.info(f"Using afs-fallback knobs.txt as definition file: '{KNOBS_FILE_AFS}'")
```
fsoubelet, line 471: Maybe let's specify explicitely we load the `knobs.txt` from `acc-models-lhc` that is in AFS (verbose, I know)?

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+#### 2024-07-08 - v0.15.0 - _jdilly_
+
+- PINNING NUMPY TO < 2.0.0
+
+- Changed:
+  - Model creation:
+    - removed hard-coded `knobs.madx` from `lhc`
+    - removed `corrections.madx` from `lhc` best-knowledge model
+    - zip up log-output files in `response_madx.py`
+    - keep 0th output file in `response_madx.py` for reference
```
fsoubelet, line 12: maybe let's quickly say this is to have insight into the model setup

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # OMC3 Changelog
 
+#### 2024-07-08 - v0.15.0 - _jdilly_
+
+- PINNING NUMPY TO < 2.0.0
+
+- Changed:
+  - Model creation:
+    - removed hard-coded `knobs.madx` from `lhc`
+    - removed `corrections.madx` from `lhc` best-knowledge model
```
fsoubelet, line 10: let's say that it was unused

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # OMC3 Changelog
 
+#### 2024-07-08 - v0.15.0 - _jdilly_
+
+- PINNING NUMPY TO < 2.0.0
+
+- Changed:
+  - Model creation:
+    - removed hard-coded `knobs.madx` from `lhc`
```
fsoubelet, line 9: 🥳

omc3/model/accelerators/lhc.py
```diff
@@ -390,24 +390,22 @@ def _get_madx_script_info_comments(self) -> str:
             f"! Model directory: {Path(self.model_dir).absolute()}\n"
             f"! Natural Tune X         [{self.nat_tunes[0]:10.3f}]\n"
             f"! Natural Tune Y         [{self.nat_tunes[1]:10.3f}]\n"
-            f"! Best Knowledge:        [{'NO' if self.model_best_knowledge is None else 'OK':>10s}]\n"
+            f"! Best Knowledge:        [{'NO' if self.model_best_knowledge is None else 'YES':>10s}]\n"
```
fsoubelet, line 393: 👍

fsoubelet (commented):
Few comments here and there for now. Do you have an idea how much space is saved by zipping the response_creation log?

CHANGELOG.md
JoschD: It was not unused by the code. We just never used it, but we could have. But it comes from a time when you couldn't manually specify the modifiers and also there was only one... So now this is all way more flexible

omc3/knob_extractor.py
JoschD: Given that the actual path is written into the same message, it's a bit overkill.... but meh...

fsoubelet (approved):

PR #442, opened 2024-08-12
Migrate to pyproject.toml and common workflows, update for numpy 2.0

fsoubelet:
Next in the wave of PRs transitioning our packages to using the standard `pyproject.toml` file.

### Important

The `setup.py`, `setup.cfg` and `MANIFEST.in` files are gone. A `pyproject.toml` file now centralises everything, including metadata conformly to [PEP621](https://peps.python.org/pep-0621/).

I have put down `hatchling` as a build backend, as it is the one developped by `PyPA`, the [Python Packaging Authority](https://www.pypa.io/en/latest/), as they publish the **authoritative** user guide. This has little impact though, as any [PEP518](https://peps.python.org/pep-0518/)-compliant backend would work there instead.

Simple changes to have `numpy 2.0` compatibility.

Like for all other packages, this moves the minimum supported Python version to `3.9`.

### Details

I have put [Hatch](https://hatch.pypa.io/latest/) as a packaging tool as it is also the one from PyPA, but it is **strictly optional**. One does not need Hatch, because standards. The new common workflows (PR upcoming) don't even use it.

I have taken upon me to remove unused declared dependencies (`hypothesis`, `attrs` etc) and sometime up minimum required versions (`tfs-pandas` for instance), for uniformity with the rest of the packages.

### Additional

This PR also moves `omc3` to the common Github Actions workflows, and brings very minor lints. Do not let the number of modified files fool you, these are very small.


fsoubelet:
As with other PRs where migration to common workflows was done, I will update the requested passing tests when the PR is accepted.

fsoubelet:
This requires a release after merge, to have a numpy 2 compatible version out there. I have updated version and changelog.

fsoubelet:
Just like for `tfs-pandas`, we have the issue of `pytables` not still being ready for `numpy 2`. See https://github.com/pylhc/tfs/pull/131#issuecomment-2283777850. @JoschD let's discuss how to handle this PR, which is more complex.

omc3/utils/iotools.py
```diff
@@ -58,7 +58,7 @@ def copy_item(src_item, dest):
         LOG.error(f"Could not copy item because of IOError. Item: '{src_item}'")
 
 
-def glob_regex(path: Path, pattern: str) -> "filter object":
+def glob_regex(path: Path, pattern: str) -> Sequence[str]:
```
JoschD, line 61: I don't think this is correct. See e.g. https://stackoverflow.com/q/33174276 or test it yourself.
On the other hand, this also regularly is marked by my linter/syntax checker as an error...

pyproject.toml
```diff
@@ -0,0 +1,114 @@
...
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.version]
+path = "omc3/__init__.py"
+
+[tool.hatch.build.targets.sdist]
+include = [
+  "/omc3",
+]
```
JoschD, line 11: why needed?

pyproject.toml
```diff
@@ -0,0 +1,114 @@
...
+  "matplotlib >= 3.8",
+  "Pillow >= 6.2.2",  # not our dependency but older versions crash with mpl
+  "generic-parser >= 1.1",
+  "sdds >= 0.4",
+  "optics-functions >= 0.1",
+  "turn_by_turn >= 0.6",
+  "uncertainties >= 3.1",
+  "scikit-learn >= 1.0",
+  "h5py >= 2.9",
+  "tables >= 3.6",
```
JoschD, line 64: anything we can do about this?

pyproject.toml
```diff
@@ -0,0 +1,114 @@
...
+optional = [
+  "pymupdf >= 1.22",  # logbook for conversion from pdf to png
+  "qtpy >= 2.3.1",  # putting plots in windows
+]
+test = [
+  "jpype1 >= 1.3",
+  "pytest >= 7.0",
+  "pytest-cov >= 2.9",
+  "pytest-timeout >= 1.4",
+  "gitpython >= 3.1",
```
JoschD, line 84: a quick search on github didn't show me that this was used anywhere in our code. Can you double-check?

JoschD (changes requested):

omc3/optics_measurements/rdt.py
```diff
@@ -61,19 +62,21 @@ def _generate_plane_rdts(order):
...
+    Computes the RDTs for the given input files and settings up to the magnet
+    order given in the inputs, and writes the results to file.
 
     Args:
-        measure_input:
-        input_files:
+        measure_input: `OpticsInput` object containing analysis settings.
+        input_files: `InputFiles` object containing frequency spectra files (linx/y).
         tunes:
         invariants:
```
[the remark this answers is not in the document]
fsoubelet: Thanks!

omc3/optics_measurements/rdt.py
```diff
@@ -187,11 +190,11 @@ def _process_rdt(meas_input, input_files, phase_data, invariants, plane, rdt):
...
     rdt_angles = stats.circular_mean(rdt_phases_per_file, period=1, axis=1) % 1
-    df[f"PHASE"] = rdt_angles
-    df[f"{ERR}PHASE"] = stats.circular_error(rdt_phases_per_file, period=1, axis=1)
+    df[f"{PHASE}"] = rdt_angles
+    df[f"{ERR}{PHASE}"] = stats.circular_error(rdt_phases_per_file, period=1, axis=1)
     df[AMPLITUDE], df[f"{ERR}{AMPLITUDE}"] = _fit_rdt_amplitudes(invariants, line_amp, plane, rdt)
-    df[f"REAL"] = np.cos(2 * np.pi * rdt_angles) * df.loc[:, AMPLITUDE].to_numpy()
-    df[f"IMAG"] = np.sin(2 * np.pi * rdt_angles) * df.loc[:, AMPLITUDE].to_numpy()
+    df[f"{REAL}"] = np.cos(2 * np.pi * rdt_angles) * df.loc[:, AMPLITUDE].to_numpy()
+    df[f"{IMAG}"] = np.sin(2 * np.pi * rdt_angles) * df.loc[:, AMPLITUDE].to_numpy()
```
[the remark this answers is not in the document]
fsoubelet: good catch

omc3/plotting/optics_measurements/constants.py
```diff
@@ -4,16 +4,7 @@
...
 """
-from dataclasses import dataclass
-from typing import Optional
-
-from omc3.optics_measurements.constants import (
-    S, ERR, DELTA, AMPLITUDE, BETA, PHASE, REAL, IMAG,
-    AMP_BETA_NAME, BETA_NAME, CHROM_BETA_NAME, PHASE_NAME,
-    SPECIAL_PHASE_NAME, TOTAL_PHASE_NAME, DISPERSION_NAME,
-    NORM_DISP_NAME, ORBIT_NAME, KICK_NAME, IP_NAME, PHASE_ADV, MDL
-)
```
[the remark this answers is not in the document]
fsoubelet: I put them back with a comment :)

pyproject.toml
fsoubelet: That would require a rework of the response matrices output, which is where we use it. Currently we have an HDF table storing the matrices but we might want a zipped folder with one .parquet per matrix for instance. 

I think it's out of this PR's scope though.

omc3/utils/iotools.py
fsoubelet: Indeed, this would be a correct type hint if we cast to `list` before returning. I'll just put a `filter` class as hint.

pyproject.toml
fsoubelet: It imports as `git` in our conftest and you introduced it as a "better" way to get `acc-models-lhc` in fixtures.

omc3/utils/iotools.py
JoschD: I'd also be ok with rewriting:

```python
def glob_regex(path: Path, pattern: str) -> Iterator[str]:
""" Do a glob on the given `path` based on the regular expression `pattern`.
    Returns only the matching filenames (as strings).
    Args:
        path (Path): Folder path to look in.
        pattern (str): Pattern to match.
    """
    regex = re.compile(pattern)
    names = (p.name for p in path.glob("*")
    return (name for name in names if regex.match(name))
 ```

pyproject.toml
JoschD: grrr... I even looked in `conftest` but didn't see it. And looked in model creation, as I assumed it had something to do with the acc-models... thanks for finding it!

Can you write `# imported as git` behind, for future reference? ^^

pyproject.toml
JoschD: I meant mostly for the tests and for allowing numpy 2.0 for now

fsoubelet: I have already excluded the relevant tests from the run, but the problem is things fail at import time. As soon as we import omc3 somewhere down the line there is an `import tables` statement and that crashes the run :(

fsoubelet: Setting numpy<2 for now

JoschD (dismissed):

fsoubelet:
## Status

@JoschD 

Same thing as for tfs-pandas (see: https://github.com/pylhc/tfs/pull/131#issuecomment-2286665125).

JoschD (approved):

fsoubelet:
Changed to make a patch release since thanks to pytables we can have full numpy 2 compatibility on Python 3.10+ at least. Will release after merge.

PR #444, opened 2024-08-20
Rescale error only if present

JoschD:
@fscarlier can you test if this helps?


JoschD:
Sometimes I wonder why I try to do things for you, @fscarlier  😢

-> Moved to another brach

Issue #457, opened 2024-09-16
[Bug]: Correctors in .json file not up-to-date


JoschD:
### Bug Description

Some of the correctors that are in the variable groups in the .json file are nowadays not allowed to be used. They should be removed from the main category that we are using and the current set should be saved for posterity under a new name.

Details to be discussed in the OMC meeting on 17.09.2024, but needs to be ready for pp-ref commissioning on the 18.09.2024

Also: MQY_Q4 exists twice (at least in Beam 1)

JoschD:
Fixed in #458

Issue #448, opened 2024-09-04
[Bug]: Global Correction: OMP not working


JoschD:
### Bug Description

The OMP method in the global corrections is not working.

- [x] write test
- [x] fix

JoschD:
Fixed in #458

Issue #447, opened 2024-08-23
[Bug]: Corrections Check window does not open, because beta_phase_y runs into stop-iteration


JoschD:
### Bug Description

Corrections Check plotting window does not open, because `beta_phase_y` runs into a stop-iteration.

Reproduce with measurement from
2024-08-22/LHCB1/Results/LHCB1_FODO_60_inj_on_off_mom

JoschD:
After testing this, it turned out to be an issue with `normalized_dispersion_` in this case, and any single-plane file in general (of which we only have normalized dispersion so far): The plot window always expected two files to be present.

JoschD:
Fixed in #458

Issue #446, opened 2024-08-23
[Bug]: Variables seem to be taken without check


JoschD:
### Bug Description

When calling global correction with variable group names that do not exist, this is ignored. (is there a warning?)
~also MQY_Q4 exists twice (at least in Beam 1)~[moved to #457]

JoschD:
There was a check, but it was logged to debug. Now `info`.

JoschD:
Fixed in #458

PR #458, opened 2024-09-16
Global Correction Fixes

JoschD:
- [x] #457 
- [x] #446 
- [x] #447 
- [x] #448


JoschD:
I open this up for review, as the only thing missing would be some changes to the correction-jsons as to be decided tomorrow morning.

omc3/correction/filters.py
```diff
@@ -27,14 +29,17 @@
 from omc3.optics_measurements.constants import AMPLITUDE, F1001, F1010, IMAG, REAL
```
fsoubelet, line 29: We could merge these two imports

omc3/model/accelerators/lhc.py
```diff
@@ -218,21 +218,25 @@ def get_variables(self, frm: float = None, to: float = None, classes: Iterable[s
...
+        if classes is not None:
+            known_classes = [c for c in classes if c in all_vars_by_class]
+            unknown_classes = [c for c  in classes if c not in all_vars_by_class]
+
+            add_knobs = set(knob for knob in unknown_classes if knob[0] != "-")
+            if add_knobs:
+                LOGGER.info("The following names are not found as corrector/variable classes and "
+                            f"are assumed to be the variable names directly instead:\n{str(add_knobs)}")
+
+            remove_knobs = set(knob[1:] for knob in unknown_classes if knob[0] == "-")
```
fsoubelet, line 231: Would not mind a comment line above here saying "if the correction variable name is prepended with '-' the it is taken out of the ones we use"

tests/unit/test_global_correction.py
```diff
@@ -58,3 +60,49 @@ def test_read_measurement_data(tmp_path, model_inj_beams, orientation):
...
+
+    correction = tfs.read(tmp_path / f"{CORRECTION_DEFAULTS['output_filename']}.tfs", index="NAME")
+    assert not np.any(correction["DELTA"].isna())
+    assert not correction["DELTA"].isin([np.inf, -np.inf]).any()
+
+    if method == "omp":
+        assert len(correction.index) == n_correctors
+
+    if method == "pinv":
+        assert len(correction.index) > n_correctors  # unless by accident
```
fsoubelet, line 108: We are not checking that the correction "works" here. It is done in the accuracy tests, right? It's late and I'm lazy to check, if yes feel free to just resolve this thread.

fsoubelet (approved):
Did I understand correctly that the OMP fix is just because `sklearn` started demanding named arguments, right?

tests/unit/test_global_correction.py
JoschD: exactly

JoschD:
> Did I understand correctly that the OMP fix is just because `sklearn` started demanding named arguments, right?

Yes. The `OrthogonalMatchingPursuit` initializor is now

```python 
def __init__(
        self,
        *,
        n_nonzero_coefs=None,
        tol=None,
        fit_intercept=True,
        precompute="auto",
    ):
```

omc3/model/accelerators/lhc.py
```diff
@@ -212,32 +214,41 @@ def get_lhc_error_dir() -> Path:
...
-        correctors_dir = LHC_DIR / "2012" / "correctors"
-        all_vars_by_class = _merge_jsons(
-            correctors_dir / f"correctors_b{self.beam}" / "beta_correctors.json",
-            correctors_dir / f"correctors_b{self.beam}" / "coupling_correctors.json",
-            self._get_triplet_correctors_file(),
+        corrector_beam_dir = Path(f"correctors_b{self.beam}")
+        all_vars_by_class = _load_jsons(
+            *self._get_corrector_files(corrector_beam_dir / "beta_correctors.json"),
+            *self._get_corrector_files(corrector_beam_dir / "coupling_correctors.json"),
+            *self._get_corrector_files("triplet_correctors.json"),
```
fsoubelet, line 221: So with the changes below, we now look for the corrector files that were put in the model directory and default to the hard-coded ones included with `omc3`, right?

JoschD: Almost. 
Up until now we had all the `corrector.json` files in the `models/lhc/2012/correctors` dir. And then adapted these files as we pleased.  I did not like this for two reasons:
a) The files were becoming really big
b) If we change the correctors to use at some point, e.g. remove knobs in 2024 from the  "MQM" category, because we cannot use them anymore from now on, calculating corrections for older models, e.g. 2018, become hard to trace back, because back in 2018 we used different knobs that are now present.

So now, I moved the `correctors` dir from 2012 directly to `models/lhc/correctors` and designated as "default".
In case we want to change the categories or add new ones that are specific for a year, we can place them in "year/correctors". 
Both files, default and year, will always be loaded, but because the `.json` file from the year is loaded later, it has priority in case it defines the same category name.

JoschD: That being said, I do like your misunderstanding... I could easily add a third layer, that first checks in the model directory, making it super flexible for online-analysis and avoiding the hacking at all. 
That was also a reason, why I implemented giving the knob names directly and also be able to remove them, because often the "hacks" are only small changes.

omc3/model/accelerators/lhc.py
```diff
@@ -519,22 +519,43 @@ def _get_corrector_elems(self) -> Path:
...
-        specific_file = self._get_accel_file(Path(Lhc.DEFAULT_CORRECTORS_DIR.name) / file_name)
+        if not default_file.exists():
+            msg = (f"Could not find {file_name} in {Lhc.DEFAULT_CORRECTORS_DIR}."
+                  "Something went wrong with the variables getting logic.")
+            raise FileNotFoundError(msg)
+        
+        LOGGER.debug(
+            f"No corrector file found for {file_name} in {default_file.parent}."
+        )
+        corrector_files = [default_file]
```
JoschD, line 539: sorry, my bad. Should read "file found"

fsoubelet (approved):
Thanks for the comment linking to our discussion 🙏

PR #460, opened 2024-09-18
Fix Various Warnings

fsoubelet:
Looking at the outputs of our tests, we have a variety of warnings that we might or might not want to care about. This PR makes changes so that the warnings disappear. See the changelog file for a list.

**Note**: some of these warnings we might want to keep, which should be a matter of revert one or a few commits. I'll create comment threads pointing at various changes and what they adress.

There remains one `RuntimeWarning: invalid value encountered in scalar multiply` emitted due to `omc3/kmod/analysis.py:333` which I haven't been able to fix.


omc3/correction/response_io.py
```diff
@@ -4,104 +4,134 @@
...
+    # If encountering issues, remove the context manager and debug
+    with ignore_natural_name_warning():
+        with pd.HDFStore(path, mode="r") as store:
+            _check_keys(store, optics_parameters, "fullresponse")
+
+            fullresponse = defaultdict(pd.DataFrame)
+            if optics_parameters is None:
+                optics_parameters = _main_store_groups(store)
+            for param in optics_parameters:
+                fullresponse[param] = store[param]
```
fsoubelet, line 47: This addresses `NaturalNameWarning`s happening during HDF5 tables operations. This is because we have keys such as `kq4.r8b2` which pytables wants to raise awareness about (we can't do `table.kq4.r8b2` as this resolves as two attributes accesses).

It is not avoidable due to our variables, and not a problem since `pandas` properly handles access operations for us (via `getattr`).

fsoubelet: Same for other places in this file which call this context manager.

omc3/correction/response_io.py
```diff
@@ -4,104 +4,134 @@
...
+    emitted within, which is our case comes from pytables. It warns about
+    table entries such as 'kq4.r8b2' which we can't access with syntax such
+    as some_table.kq4.r8b2 but we don't care about this. We let pandas handle
+    the access with getattr (which works).
+
+    If encountering issues, comment out the context manager and debug.
+    """
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", NaturalNameWarning)
+        yield
```
fsoubelet, line 137: Simply takes care of filtering out the `NaturalNameWarning`.

omc3/kmod/analysis.py
```diff
@@ -217,15 +220,18 @@ def do_fit(magnet_df, plane, use_approx=False):
...
+        warnings.simplefilter("ignore", category=OptimizeWarning)
+        av_beta, av_beta_err = scipy.optimize.curve_fit(
+            fun,
+            xdata=return_fit_input(magnet_df, plane),
+            ydata=magnet_df.where(magnet_df[f"{CLEANED}{plane}"])[
+                f"{TUNE}{plane}"].dropna() - magnet_df.headers[f"{TUNE}{plane}"],
+            sigma=sigma,
+            absolute_sigma=True,
+            p0=1
+        )
```
fsoubelet, line 234: **Note**: I have filtered the `OptimizeWarning: Covariance of the parameters could not be estimated` one as it feels like we can't do much about it, but we might decide to keep them.

omc3/optics_measurements/rdt.py
```diff
@@ -267,10 +268,14 @@ def fitting(x, f):
...
-        sqrt_pcov = np.sqrt(pcov).flat[0]
-        err_amps[i] = sqrt_pcov if np.isfinite(sqrt_pcov) else 0. # if single file is used, the error is reported as Inf, which is then overwritten with 0
+        # Catch the potential 'OptimizeWarning: Covariance of the parameters could not be estimated'
+        # that can happen during the curve_fit
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", OptimizeWarning)
+            popt, pcov = curve_fit(fitting, kick_data, bpm_rdt_data, p0=guess[i])
+            amps[i] = popt[0]
+            sqrt_pcov = np.sqrt(pcov).flat[0]
+            err_amps[i] = sqrt_pcov if np.isfinite(sqrt_pcov) else 0. # if single file is used, the error is reported as Inf, which is then overwritten with 0
```
fsoubelet, line 278: **Note**: Same here, we might want to keep.

omc3/plotting/plot_amplitude_detuning.py
```diff
@@ -521,9 +526,9 @@ def _plot_detuning_3d(ax: Axes, data: Dict[str, AmpDetData], label: str, color=N
 
     # plot errorbars
     for idx in np.arange(0, len(tune)):
-        x, dx = jx[idx], jx_err[idx]
-        y, dy = jy[idx], jy_err[idx]
-        z, dz = tune[idx], tune_err[idx]
+        x, dx = jx.iloc[idx], jx_err.iloc[idx]
+        y, dy = jy.iloc[idx], jy_err.iloc[idx]
+        z, dz = tune.iloc[idx], tune_err.iloc[idx]
```
fsoubelet, line 531: This fixes the `FutureWarning` for using `Series.__getitem__` which is deprecated.

omc3/plotting/plot_amplitude_detuning.py
```diff
@@ -648,9 +653,14 @@ def _format_axes_3d(
...
-    ax.figure.tight_layout()
-    ax.figure.tight_layout()
+
+    # tight layout so that the legend fits in figure
+    # We catch and ignore 'UserWarning: The figure layout has changed to tight'
+    # because it is something we did on purpose, let's not pollute the output.
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", UserWarning)
+        ax.figure.tight_layout()
+        ax.figure.tight_layout()
```
fsoubelet, line 663: This I think is reasonable. It filters out the `UserWarning` matplotlib issues to notify the figure layout has changed to tight as this is something we explicitely ask for.

omc3/plotting/spectrum/waterfall.py
```diff
@@ -87,11 +87,15 @@ def _format_axes(fig_cont, limits, ncol):
...
         else:
-            ax.set_yticklabels(ylabels, fontdict={'fontsize': get_fontsize_as_float(matplotlib.rcParams[u'axes.labelsize']) * .5})
-            ax.set_yticks(range(len(ylabels)))
+            # Provide ticks and labels together or matplotlib issues a UserWarning
+            # See "Discouraged" admonition at https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set_yticklabels.html 
+            ax.set_yticks(
+                ticks=range(len(ylabels)),
+                labels=ylabels,
+                fontdict={'fontsize': get_fontsize_as_float(matplotlib.rcParams[u'axes.labelsize']) * .5},
+            )
```
fsoubelet, line 98: This fixes the `UserWarning: set_ticklabels() should only be used with a fixed number of ticks, i.e. after set_ticks() or using a FixedLocator.`

The recommended way according to the `matplotlib` documentation is to use `axis.set_ticks` and provide both ticks and labels as the same time.

omc3/plotting/utils/annotations.py
```diff
@@ -298,17 +305,31 @@ def reorder(items):
...
+        leg = ax.legend(
+            handles=handles,
+            labels=labels,
+            loc="lower right",
+            bbox_to_anchor=(1.0, 1.0+pad),
+            fancybox=frame,
+            shadow=frame,
+            frameon=frame,
+            ncol=ncol,
+        )
```
fsoubelet, line 332: **Note**: Do we want to keep these showing up?

This relates to the `UserWarning: No artists with labels found to put in legend.  Note that artists whose label start with an underscore are ignored when legend() is called with no argument.`.

I assumed most of the time this happens because there is a leading underscore, or we didn't write a label for every / any lines. We might want to be more thorough and write those every time instead of filtering the warning here.

omc3/response_creator.py
```diff
@@ -76,40 +77,43 @@
...
+        type=str,
+        nargs="+",
+        choices=OPTICS_PARAMS_CHOICES,
+        help="List of parameters to correct upon (e.g. BBX BBY; twiss-only).",
+    )
+    params.add_parameter(
+        help="Print debug information.",
+        name="debug",
+        action="store_true",
+    )
```
fsoubelet, line 111: Formatter triggered, sorry.

omc3/run_kmod.py
```diff
@@ -285,8 +285,11 @@ def create_lsa_results_file(betastar_required, instruments_found, results_df, in
         exporting_columns=['LABEL', f'{BETA}{STAR}X', f'{ERR}{BETA}{STAR}X', f'{BETA}{STAR}Y', f'{ERR}{BETA}{STAR}Y']
         lsa_results_df=results_df[exporting_columns].rename(columns=dict(zip(exporting_columns, LSA_COLUMNS)))
     if instruments_found:
-        lsa_results_df = pd.concat([lsa_results_df, instrument_beta_df],
-                                   axis="index", sort=False, ignore_index=True)
+        # We first make sure we don't try a concat operation if a df is empty
+        # (otherwise pandas complains with a FutureWarning since 2.1.1)
+        dfs_to_concat = [lsa_results_df, instrument_beta_df]
+        dfs_to_concat = list(filter(lambda df: not df.empty, dfs_to_concat))
+        lsa_results_df = pd.concat(dfs_to_concat, axis="index", sort=False, ignore_index=True)
```
fsoubelet, line 292: This fixes the `FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.`

I added a simple check where only non-empty dataframes from these two are passes to the concatenation.

omc3/scripts/betabeatsrc_output_converter.py
```diff
@@ -89,8 +88,8 @@ def converter_entrypoint(opt: DotDict) -> None:
     """
     iotools.create_dirs(Path(opt.outputdir))
     save_options_to_config(
-        Path(opt.outputdir) / DEFAULT_CONFIG_FILENAME.format(time=datetime.utcnow().strftime(formats.TIME)),
-        OrderedDict(sorted(opt.items())),
+        Path(opt.outputdir) / DEFAULT_CONFIG_FILENAME.format(time=datetime.now(timezone.utc).strftime(formats.TIME)),
+        dict(sorted(opt.items())),
```
fsoubelet, line 92: datetime again

omc3/tbt_converter.py
```diff
@@ -107,7 +107,7 @@ def converter_entrypoint(opt):
     save_options_to_config(
         str(
             Path(opt.outputdir)
-            / DEFAULT_CONFIG_FILENAME.format(time=datetime.utcnow().strftime(formats.TIME))
+            / DEFAULT_CONFIG_FILENAME.format(time=datetime.now(timezone.utc).strftime(formats.TIME))
```
fsoubelet, line 110: Same

omc3/utils/outliers.py
```diff
@@ -113,8 +113,16 @@ def get_filter_mask(data: ArrayLike, x_data: ArrayLike = None, limit: float = 0.
...
+    import warnings
+    try:
+        from numpy.exceptions import RankWarning as IgnoredWarning
+    except ImportError:  # it does not exist on numpy versions supporting 3.9
+        IgnoredWarning = Warning
+    # We filter out the "Polyfit may be poorly conditioned" warning
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=IgnoredWarning)
+        m, b = np.polyfit(x[mask], y[mask], 1)
+        return y[mask] - b - m * x[mask], y - b - m * x
```
fsoubelet, line 125: This is similar to the `scipy` `OptimizeWarning` in another file, but ignores the `RankWarning` from `numpy` for the same reason, there isn't much we can do about it.

Caveat is that this warning only appeared in the first `numpy` version to not support Python 3.9 so some of our CI workflows would fail here. I add a hacky check.

**Note**: We might decide to keep this.

tests/accuracy/twiss_to_lin.py
```diff
@@ -119,5 +118,5 @@ def _get_header(tunes, nattunes, plane):
         f"Q{PLANE_TO_NUM[plane]}RMS": 1e-7,
         f"NATQ{PLANE_TO_NUM[plane]}": nattunes[plane],
         f"NATQ{PLANE_TO_NUM[plane]}RMS": 1e-6,
-        "TIME": datetime.utcnow().strftime(formats.TIME),
+        "TIME": datetime.now(timezone.utc).strftime(formats.TIME),
```
fsoubelet, line 121: datetime again

fsoubelet:
I am also adding one or two things from an old maintenance branch:

- Display of coverage stats in the `coverage` workflow.
-

omc3/kmod/analysis.py
JoschD: I would leave it in, as it gives the user some feedback that there might be an issue and how much to trust the fit!

omc3/optics_measurements/rdt.py
JoschD: same comment

omc3/plotting/utils/annotations.py
JoschD: I think this is a reasonable warning, and should be taken care of in the test that raises it. I.e. assign some labels or do not create the legend.

omc3/kmod/analysis.py
fsoubelet: In this case, I would like to catch the warnings anyway but emit it as a log message ourselves. This way we are not polluting `stderr` and the user can choose to see this or not by setting the logging level. Thoughts?

CHANGELOG.md
```diff
@@ -1,6 +1,23 @@
 # OMC3 Changelog
 
-#### 2024-09-18 - v0.16.0 - _jdilly_
+## IN PROGRESS - v0.16.1 - _fsoubelet_
+
```
JoschD, line 4: I left it #### in the past because I like the format of how it looks better ... but if you prefer it this way, or are more bothered by the IDE warning than how the log looks like, this is also ok

fsoubelet: No strong opinions, I tend to make the linters happy but it's not like it's a big deal, it's barely code.

omc3/response_creator.py
```diff
@@ -76,40 +77,43 @@
...
+        default="madx",
+        help="Create either with madx or analytically from twiss file.",
+    )
+    params.add_parameter(
+        name="variable_categories",
+        nargs="+",
+        default=CORRECTION_DEFAULTS["variable_categories"],
+        help="List of the variables classes to use.",
+    )
+    params.add_parameter(name="outfile_path", type=PathOrStr, help="Name of fullresponse file.")
```
JoschD, line 93: at least expand this one then as well ^^

omc3/run_kmod.py
JoschD: I think list comprehension in this case is more readable and probably faster (not that this matters).
```
dfs_to_concat = [df for df in dfs_to_concat if not df.empty()]
```

But then also:
```
if not len(dfs_to_concat):
    msg = "All dfs are empty! Something went wrong along the way."
    raise ValueError(msg)
 ```

Optional
```
if len(dfs_to_concat) == 1:
    lsa_results_df = dfs_to_concat[0]
else:
    ...
```

File: omc3/utils/outliers.py (hunk 13 above)
Comment (JoschD) [reply omc3/utils/outliers.py:125]: catch and log as well

**Review by JoschD — CHANGES_REQUESTED** [review]

File: omc3/plotting/utils/annotations.py (hunk 8 above)
Comment (fsoubelet) [reply omc3/plotting/utils/annotations.py:332]: I reverted the intercept. Did not manage to figure out where in the tests we were encountering this specific condition, so for now it will stay in the outputs.

**Review by JoschD — CHANGES_REQUESTED** [review]

File: omc3/plotting/utils/annotations.py [hunk 17]
```diff
@@ -4,17 +4,23 @@
 
 Helper functions to create annotations, legends as well as style labels in plots.
 """
+from __future__ import annotations
+
 import itertools
 import re
-from typing import Union
+from typing import TYPE_CHECKING
```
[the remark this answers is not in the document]
JoschD: Ignore, it was just not in the changes somehow

omc3/plotting/utils/annotations.py
```diff
@@ -298,17 +304,29 @@ def reorder(items):
...
 
     if transposed:
         handles, labels = transpose_legend_order(ncol, handles, labels, ax)
 
-    leg = ax.legend(handles=handles, labels=labels, loc='lower right',
-                    bbox_to_anchor=(1.0, 1.0+pad),
-                    fancybox=frame, shadow=frame, frameon=frame, ncol=ncol)
+    # It is possible that no labels are given to the legend commands,
+    # or they are given with a leading underscore, which will get them
+    # ignored. In this case matplotlib issues a UserWarning here
```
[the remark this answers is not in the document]
fsoubelet: Removed.

omc3/response_creator.py
```diff
@@ -76,40 +77,47 @@
...
+        help="Print debug information.",
+        name="debug",
+        action="store_true",
+    )
     return params
 
 
 @entrypoint(response_params())
-def create_response_entrypoint(opt: DotDict, other_opt) -> Dict[str, pd.DataFrame]:
+def create_response_entrypoint(opt: dict, other_opt) -> dict[str, pd.DataFrame]:
```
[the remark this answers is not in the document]
fsoubelet: Don't think it matters much but I put it back.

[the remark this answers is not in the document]
fsoubelet: Put it back.

JoschD (approved):
Nice fix! I am only wondering how many more warnings will pop up soon... I already thought I fixed most of them in `0.15.3` but there were now so many more ^^

PR #461, opened 2024-09-18
new parameter: analyse_dpp

JoschD:
Let the user choose if they want to analyse all files or only files with specific dpp (most likely the on-momentum files)


omc3/optics_measurements/phase.py
```diff
@@ -48,35 +83,48 @@ def calculate(
...
-
-    if len(phase_advances["MEAS"].index) < 3:
+        LOGGER.info("-- Run on driven model")
+
+        # TODO:Verify this is correct (that's how it was implemented before)
+        driven_compensation = (
+            CompensationMode.EQUATION 
+            if meas_input.compensation == CompensationMode.EQUATION else 
+            CompensationMode.NONE
+        )
```
JoschD, line 113: @lmalina @emaclean @awegsche @rogeliotomas 
I tried to make this a bit clearer how the logic is here, as this was hidden within the function. But I am not sure if this is correct as it is:
The way it had been implemented is, that the equation compensation also operates on the driven data, but the model compensation only on the free data, if I see that correctly.

Could you please confirm that this is correct? My knowledge of these compensations is very limited.

omc3/optics_measurements/phase.py
```diff
@@ -42,6 +50,15 @@
...
 LOGGER = logging_tools.get_logger(__name__)
 
+class CompensationMode:
+    NONE: str = "none"
+    MODEL: str = "model"
+    EQUATION: str = "equation"
+    
+    @classmethod
+    def all(cls) -> list[str]:
+        return [cls.NONE, cls.MODEL, cls.EQUATION]
```
fsoubelet, line 60: Do we want to make this an `Enum`? I'm not sure I understand well how they work though

omc3/optics_measurements/phase.py
```diff
@@ -151,33 +183,37 @@ def _calculate_with_compensation(
...
-    df = pd.merge(df, input_files.joined_frame(plane, [f"MU{plane}", f"{ERR}MU{plane}"],
-                                               dpp_value=dpp_value, how=how),
-                  how='inner', left_index=True, right_index=True)
-    df[input_files.get_columns(df, f"MU{plane}")] = (input_files.get_data(df, f"MU{plane}")
-                                                     * meas_input.accelerator.beam_direction
-                                                     )
-    phases_mdl = df.loc[:, f"MU{plane}"].to_numpy()
-    phase_advances = {"MODEL": _get_square_data_frame(_get_all_phase_diff(phases_mdl), df.index)}
-    if compensation == "model":
+    joined_df = input_files.joined_frame(plane, [f"{PHASE_ADV}{plane}", f"{ERR}{PHASE_ADV}{plane}"], dpp_value=dpp_value, how=how)
```
fsoubelet, line 193: It was never so clear to me what exactly comes out of `input_files.joined_frame` (docstrings there did not help). Could we add a comment or change the variable name to make this a bit more explicit?

omc3/optics_measurements/phase.py
```diff
@@ -105,14 +137,13 @@ def calculate(
...
 def _calculate_with_compensation(
     meas_input: DotDict,
     input_files: InputFiles,
     tunes: TuneDict,
     plane: str,
     model_df: pd.DataFrame,
-    compensation: str = "none",
+    compensation: str = CompensationMode.NONE,
     no_errors: bool = False,
 ):
```
fsoubelet, line 148: Let's type hint the return type too

tests/unit/test_hole_in_one.py
```diff
@@ -75,30 +74,39 @@ def test_hole_in_two(tmp_path, clean, which_files, dpp, caplog):
...
-        nonlinear=['rdt', 'crdt'],
-        outputdir=optics_output,
-    )
-
-    _check_linear_optics_files(optics_output, off_momentum=(which_files == "all"))
-    _check_nonlinear_optics_files(optics_output, "rdt")
-    _check_nonlinear_optics_files(optics_output, "crdt")
-
-    _check_caplog_for_rdt_warnings(caplog, to_be_found=(which_files == "all") and (dpp is None))
+    for dpp in (None, 0):
```
fsoubelet, line 77: Why not as a fixture anymore?

omc3/optics_measurements/phase.py
JoschD: Not the biggest fan of enums (in pyhton). Also: As these are input-parameters, we don't want to confuse people with `--compensation 1` instead of `--compensation model` .

omc3/optics_measurements/phase.py
JoschD: added more info to the docstring of `joined_frame`. 

To be honest, it would not have been my first choice in how to implement this, but it does its job and would be a lot of work to rethink and rewrite

tests/unit/test_hole_in_one.py
JoschD: I would prefer it as fixture, but then we run also the harpy part 6 times, which does not change. On the other hand, I think the optics-measurement part is the one that takes the longest anyway...

JoschD: The new tests actually add 20min to the total test times (extended tests seem to increase, 20min -> 30min, coverage test goes from 40min -> 58min). To be seen if we want this ....

fsoubelet: Big oof...

omc3/optics_measurements/phase.py
JoschD: After some discussion and feedback from lukas, we are both relatively sure this is a bug. I will try to device a test

JoschD:
Putting on hold until I have the time to device a test to check the output of compensated and uncompensated phase against each other (or the model)

JoschD:
Priority increased to "Normal" as this is very useful for injectors

omc3/optics_measurements/data_models.py
```diff
@@ -219,3 +229,40 @@ def get_data(frame, column) -> np.ndarray:
...
+    msg += (
+        ", which can make the results more inaccurate.\n"
+        "To avoid, specify `analyse_dpp` or run the analysis only on on-momentum files"
+    )
+    
+    if id_ is not None:
+        msg += f" or possibly deactivate the {id_}"
+    
+    msg += "."
+    LOGGER.warning(msg)
```
fsoubelet, line 263: Cool

omc3/optics_measurements/crdt.py
```diff
@@ -57,11 +73,12 @@
 def calculate(measure_input: DotDict, input_files: InputFiles, invariants, header):
     """ Calculate the CRDT values. """
     LOGGER.info("Start of CRDT analysis")
-    
-    # Todo: only on-momentum required? If not, remove this or set `dpp_value=None`, see https://github.com/pylhc/omc3/issues/456
-    # For now: use only the actions belonging to the current dpp value!
-    dpp_value = 0  
-    invariants = {plane: inv[input_files.dpp_frames_indices(plane, dpp_value)] for plane, inv in invariants.items()}  
+    dpp_value = measure_input.analyse_dpp
```
fsoubelet, line 76: Do we default this value to 0?

omc3/optics_measurements/measure_optics.py
```diff
@@ -102,31 +132,36 @@ def chromatic_beating(input_files: InputFiles, measure_input: DotDict, tune_dict
...
     calibs = {}
+    calibrationdir = Path(calibrationdir)
+    outputdir = Path(outputdir)
+
     for plane in PLANES:
         cal_file = CALIBRATION_FILE.format(plane=plane.lower())
-        iotools.copy_item(os.path.join(calibrationdir, cal_file), os.path.join(outputdir, cal_file))
-        calibs[plane] = tfs.read(os.path.join(outputdir, cal_file)).set_index(NAME)
+        iotools.copy_item(calibrationdir / cal_file, outputdir / cal_file)
+        calibs[plane] = tfs.read(outputdir / cal_file, index=NAME)
```
fsoubelet, line 166: <3

fsoubelet (commented):
We should make PRs just for maintenance and flag the merge commits as revisions...

Minor questions but ok with me, we'll have to see about the order in which we merge the active PRs to resolve conflicts.

omc3/optics_measurements/measure_optics.py
```diff
@@ -102,31 +132,36 @@ def chromatic_beating(input_files: InputFiles, measure_input: DotDict, tune_dict
...
+    return {
+        "Measure_optics:version": VERSION,
+        "Command": f"{sys.executable} {' '.join(sys.argv)}",
+        "CWD": Path.cwd().absolute(),
+        "Date": datetime.datetime.today().strftime("%d. %B %Y, %H:%M:%S"),
+        "Model_directory": meas_input.accelerator.model_dir,
+        "Compensation": meas_input.compensation,
+        "Q1": tune_dict["X"]["QF"],
+        "Q2": tune_dict["Y"]["QF"],
+    }
```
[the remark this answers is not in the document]
JoschD: A pre keep-order-by-default dict way with lists.

PR #464, opened 2024-09-30
Fix knob extractor

JoschD:
While we removed `/mcr/bin` from the lhcop `PATH`, which allows us to run the knob-creator in the CCC terminals,
it is still added when starting the CCM (see https://cern.service-now.com/service-portal?id=ticket&table=incident&n=INC3768823 ). 

This PR provides a hack to remove it again ...

### DIFF
--- a/omc3/knob_extractor.py
+++ b/omc3/knob_extractor.py
@@ -283,7 +283,9 @@ def get_params():
 )
 def main(opt) -> tfs.TfsDataFrame:
     """ Main knob extracting function. """
-    ldb = pytimber.LoggingDB(source="nxcals", loglevel=logging.ERROR)
+    ldb = pytimber.LoggingDB(source="nxcals", loglevel=logging.ERROR, 
+                             sparkprops={"spark.ui.showConsoleProgress", "false"}
+    )
     time = _parse_time(opt.time, opt.timedelta)
 
     if opt.state:
--- a/omc3/knob_extractor.py
+++ b/omc3/knob_extractor.py
@@ -61,6 +61,19 @@ The data is fetched from ``NXCALS`` through ``pytimber`` using the **StateTracke
 
 
 """
+
+####### WORKAROUND FOR JAVA ISSUES WITH LHCOP ##################################
+# Move `/mcr/bin` to the end of the Path.
+# This ia a hack, please remove at the earliers convenience. Fixes:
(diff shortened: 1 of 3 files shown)


JoschD:
This PR is on hold until @tpersson tests this in the CCC. I have put this branch in `python_edge` for now. So simply creating a model from the omc3-GUI (started from the CCM) with any online-model extraction (xing, offset, disp..) enabled should do the trick.

Success or failure be very visible in the log-output (in the extended log-viewer in the GUI).

To be sure, one could switch the python to the omc_python3 in the same GUI and see that it fails.

JoschD:
Tested in the CCC and works. One needs to have a valid kerberos token,  but that has always been the case

omc3/knob_extractor.py
```diff
@@ -263,7 +283,9 @@ def get_params():
 )
 def main(opt) -> tfs.TfsDataFrame:
     """ Main knob extracting function. """
-    ldb = pytimber.LoggingDB(source="nxcals", loglevel=logging.ERROR)
+    ldb = pytimber.LoggingDB(source="nxcals", loglevel=logging.ERROR, 
+                             sparkprops={"spark.ui.showConsoleProgress", "false"}
```
fsoubelet, line 287: This is to remove `Spark`'s progress bar, right?

omc3/knob_extractor.py
```diff
@@ -61,13 +61,30 @@
...
+#
+# This ia a hack, please remove at the earliers convenience. For updates see: 
+# https://cern.service-now.com/service-portal?id=ticket&table=incident&n=INC3768823
+import os
+if "PATH" in os.environ and "/mcr/bin" in os.environ["PATH"]:
+    parts = os.environ["PATH"].split(":")
+    parts.remove("/mcr/bin")
+    parts.append("/mcr/bin")  # probably not neccesary
+    os.environ["PATH"] = ":".join(parts)
+################################################################################
```
fsoubelet, line 79: Cool that this works, annoying that we have to do it. Feel free to leave a `TODO` in there too, might be going under our radar after some time. Feel free to ignore too.

fsoubelet (approved):
Thank you. Let's hope we can remove this hack soon enough.

jgray-19 (approved):

PR #466, opened 2024-10-18
Add Calculation of DPP in full response and global correction

jgray-19:
This pull request introduces the handling of the variable delta p/p called, `DELTAP_NAME` (`"orbit_dpp"`), across multiple files in the codebase, allowing calculation of the delta p/p response. Then also enables calculation of this value, including using the iterative global correction response. 

### New Additions
* `omc3/check_corrections.py` - Add `variable_categories` as a parameter, thanks to the changes that had to be made to `global_correction._create_corrected_model`. Potentially cleaner to not have it as a parameter?
* `omc3/correction/handler.py` - Add `variable_categories` as a parameter, so `_create_corrected_model`can know if it needs to create a corrected model with an included `dpp`. Then also change the full response matrix creation to create the matrix around a new `dpp`. 
* `omc3/model/accelerators/accelerator.py` - Add a new function that needs to be overloaded called `get_update_deltap_script`. Which can allow the accelerator to do a twiss, correct, match, ensuring that the deltap calculations uses the correct orbit and tunes.
* `omc3/model/accelerators/lhc.py` - 
  - Update `get_update_correction_script` to add the necessary twiss, correct, match, if the user asks to calculate `dpp`
  - Add a `get_deltap_script` function, which does the twiss, correct, match. Unfortunately due to knob naming conventions, an awful if statement is included. 
* `tests/accuracy/test_global_correction.py` - Added tests to test my new additions. The create response matrix is mostly tested through the iterations of the global correction function, as the value will diverge if the response matrix is incorrect. Two main tests:
  - If the resulting calculated delta p/p is within 5% of the actual value that is simulated (or 2% after two iterations)
  - If the delta p/p actually converges as the number of iterations increase. Independent of the value of update_response.

### Bug Fixes and Improvements:

* [`omc3/correction/filters.py`](diffhunk://#diff-6c3b4496602badb6885e55fca937529fc78e0cbb41021ea0f319b8b033cc1f66L170-R170): Changed `LOG.warn` to `LOG.warning` as the former is depreciated.
* [`omc3/correction/handler.py`](diffhunk://#diff-1022807905606346c3484a408d137bc141402d7c5f890834dedff4ac6dea5efcL317-R328): Fixed typos in debug and info log messages.
* [`omc3/global_correction.py`](diffhunk://#diff-27b7844610b1f25fea222d01078744373e33eec3c54edb13b5450330797b77f6L19-R19): Corrected a typo in the docstring.
* [`omc3/correction/response_twiss.py`](diffhunk://#diff-402df306c09bdd1b3d02d53d7fff6c9d204748e6cfebdf031cc73e6e08883deaL120-R120): Simplified import statements and updated type hints. [[1]](diffhunk://#diff-402df306c09bdd1b3d02d53d7fff6c9d204748e6cfebdf031cc73e6e08883deaL120-R120) [[2]](diffhunk://#diff-402df306c09bdd1b3d02d53d7fff6c9d204748e6cfebdf031cc73e6e08883deaL804-R804) [[3]](diffhunk://#diff-402df306c09bdd1b3d02d53d7fff6c9d204748e6cfebdf031cc73e6e08883deaL843-R843)


omc3/check_corrections.py
```diff
@@ -462,7 +463,7 @@ def _get_measurement_filter(nominal_model: TfsDataFrame, opt: DotDict) -> dict[s
 
 def _create_model_and_write_diff_to_measurements(
         output_dir: Path, measurement: OpticsMeasurement, correction_name: str, correction_files: Sequence[Path],
-        accel_inst: Accelerator, rms_masks: dict) -> OpticsMeasurement:
+        accel_inst: Accelerator, rms_masks: dict, variable_categories: Sequence[str]) -> OpticsMeasurement:
```
JoschD, line 466: Does that make sense to add here? As far as I can see this is only called with an empty array. What is your plan for this? Adding this as optional input argument to check how well  the correction of DPP will work?

omc3/correction/handler.py
```diff
@@ -92,7 +92,18 @@ def correct(accel_inst: Accelerator, opt: DotDict) -> None:
...
+                # otherwise we go through the other way of computing the response.
+                if DELTAP_NAME in opt.variable_categories:
+                    # Update dpp and create response around this dpp, then reset dpp (So _create_corrected_model is unaffected)
+                    old_dpp = accel_inst.dpp
+                    accel_inst.dpp = delta[DELTA][DELTAP_NAME]
+                    resp_dict = response_madx.create_fullresponse(accel_inst, opt.variable_categories)
+                    accel_inst.dpp = old_dpp
+                else:
+                    resp_dict = response_twiss.create_response(accel_inst, opt.variable_categories, optics_params)
+
```
JoschD, line 106: I don't like this whole part here, the loop is already convoluted enough. Maybe extract into a function. But also a few bugs here:
- DELTAP_NAME should be checked if in vars_list, not variable_categories
- I think the `accel_inst.dpp` should be `accel_inst.dpp = accel_inst.dpp + delta` (that's why it is a delta)
- I don't like how intransparent this switch from twiss to madx response is, there should be at least a LOG.info or LOG.warning for the user

omc3/model/accelerators/lhc.py
```diff
@@ -486,12 +488,42 @@ def _get_madx_main_sequence_loading(self) -> str:
...
             )
 
-    def get_update_correction_script(self, outpath: Path | str, corr_files: Sequence[Path | str]) -> str:
+    def get_update_correction_script(self, outpath: Path | str, corr_files: Sequence[Path | str], variable_categories: Sequence[str]) -> str:
         madx_script = self.get_base_madx_script()
         for corr_file in corr_files:
             madx_script += f"call, file = '{str(corr_file)}';\n"
-        madx_script += f"exec, do_twiss_elements(LHCB{self.beam}, '{str(outpath)}', {self.dpp});\n"
+        
+        if DELTAP_NAME in variable_categories: # If the dpp is a variable in the corrections, do twiss, correct, match
```
JoschD, line 496: same comment about the `DELTAP_NAME` in the categories/variables list here. BUT as you do not need all of the variables in this function (and also hopefully never will) I would simply pass a boolean i.e. `update_dpp: bool = False` as a parameter.

omc3/model/accelerators/lhc.py
```diff
@@ -486,12 +488,42 @@ def _get_madx_main_sequence_loading(self) -> str:
...
+        else: # Else, just set the dpp and do twiss
+            madx_script += f"{DELTAP_NAME} = {self.dpp:.15e};\n"
+        madx_script += f"exec, do_twiss_elements(LHCB{self.beam}, '{str(outpath)}', {DELTAP_NAME});\n"
         return madx_script
+    
+    def get_update_deltap_script(self) -> str:
+        if self._uses_run3_macros():
+            knobx = f"dQx.b{self.beam}_op"
+            knoby = f"dQy.b{self.beam}_op"
+        elif int(self.year) == 2018: # This is very annoying - is there another way?
```
JoschD, line 507: Yes it is, but I don't think there is a better way. 
But I would move this whole block into its own function (e.g. `self.get_tune_knobs() -> tuple[str, str]:` and in this function `tune_knobs = self.get_tune_knobs()`, `"\n".join([f"vary, name={knob};" for knob in tune_knobs])`), as I easily imagine to have to use this again somewhere in the future.

omc3/model/accelerators/accelerator.py
```diff
@@ -317,7 +317,7 @@ def get_file(cls, filename: str) -> Path:
 
     # Jobs ###################################################################
 
-    def get_update_correction_script(self, outpath: Union[Path, str], corr_files: Sequence[Union[Path, str]]) -> str:
+    def get_update_correction_script(self, outpath: Union[Path, str], corr_files: Sequence[Union[Path, str]], variable_categories: Sequence[str]) -> str:
```
JoschD, line 320: see comment in `lhc.py`.
Possibly even simply `**kwargs` here?

omc3/correction/response_twiss.py
```diff
@@ -117,7 +117,7 @@
     https://cds.cern.ch/record/2632945/
 
 """
-from typing import Dict, List, Sequence
+from typing import Sequence
```
JoschD, line 120: deprecated, now `from collections.abc import Sequence`

tests/accuracy/test_global_correction.py
```diff
@@ -157,6 +159,54 @@ def test_lhc_global_correct(tmp_path, model_inj_beams, orientation):
         diff_rms_prev = diff_rms
 
 
+@pytest.mark.basic
+@pytest.mark.parametrize('dpp', (-2e-4, -1e-4, 1e-4, 7.5e-4))
```
JoschD, line 163: Are the runs different for these different dpps? If not, I would simply test one (or maybe with +-), to reduce the total testing time. Otherwise it's just like running the same test 4 times.

tests/accuracy/test_global_correction.py
```diff
@@ -195,3 +245,59 @@ def _create_fake_measurement(tmp_path, model_path, twiss_path, error_val, optics
             meas[ERROR] = meas.loc[:, f"{ERR}{col}"].to_numpy()
         meas[WEIGHT] = 1.
     return twiss_df, model_df, meas_dict
+
+# Is the following better placed in a folder with reference files,
```
JoschD, line 249: I think better to create the twiss_dpp files beforehand and simply load. 
Especially as we want to get rid of the sequence files in our repo eventually anyway.
Maybe use the acc-models-lhc way and put the creation in a script next to the twiss_dpp output files for reference and easy future recreation.

JoschD (changes requested):
klicked the wrong button. Should have been "request changes" . Sorry!

tests/accuracy/test_global_correction.py
```diff
@@ -174,8 +174,8 @@ def test_lhc_global_correct_dpp(tmp_path: Path, model_inj_beams: DotDict, dpp: f
     for key in response_dict.keys():
         assert DELTAP_NAME in response_dict[key].columns
 
-    # create and load fake measurement
-    dpp_path = run_dpp(tmp_path, dpp, model_inj_beams.beam)
+    # create fake measurement from previously created model
+    dpp_path = f"tests/inputs/correction/deltap/twiss_dpp_{dpp:.1e}_B{model_inj_beams.beam}.dat"
```
fsoubelet: Chiming in before the weekend to ask for this hard-coded path to disappear, and using `CORRECTION_INPUTS` defined at the top of file.

This way it doesn't crash when a test isn't started from the repo's root and if we change some location it'll propagate easier.

fsoubelet: Oh well, @JoschD committed that change as I was complaining :P

fsoubelet:
I like the new version (bool parameter + some pure functions) much better. Will do a review on monday morning.

tests/accuracy/test_global_correction.py
jgray-19: Yeah sorry, I committed without thinking!

omc3/correction/handler.py
```diff
@@ -109,6 +119,60 @@ def correct(accel_inst: Accelerator, opt: DotDict) -> None:
...
+    to the accelerator instance, to run the `twiss` in madx at this dpp.
+    All other parameters are taken care of in the model/elements.
+    The dpp value has to be reset afterwards, as in the calculation of the correction,
+    the delta is added via the CHANGEPARAMTERS file.
+
+    Alternatively: We could create a copy of the accelerator instance,
+    but that would require a proper copy implementation of the class, I think (jdilly 2024)
+    """
+    # update model, not nice as "private" elements, but will be reset later
+    original_model = accel_inst.model
```
jgray-19, line 148: @JoschD Could you make sure this what you meant? I got an error that accel_inst._model did not exist. (see [most recent commit](https://github.com/pylhc/omc3/pull/466/commits/a656e3777e5f8112b9ab8de3b85b99da5166e173#diff-1022807905606346c3484a408d137bc141402d7c5f890834dedff4ac6dea5efcL144-L145))

omc3/correction/handler.py
```diff
@@ -76,10 +78,14 @@ def correct(accel_inst: Accelerator, opt: DotDict) -> None:
...
         # ######### Update Model and Response ######### #
         if iteration > 0:
-            LOG.debug("Updating model via MADX.")
+            LOG.debug("Updating model via MAD-X.")
             corr_model_path = opt.output_dir / f"twiss_{iteration}{EXT}"
 
-            corr_model_elements = _create_corrected_model(corr_model_path, [opt.change_params_path], accel_inst)
+            if update_deltap:
+                # update dpp here, as it does not have a MAD-X knob.
+                update_deltap = delta.loc[ORBIT_DPP, DELTA]
```
fsoubelet, line 86: I don't like this.

We define initially `update_deltap` as a bool which is used as a flag to check if we want to do this, but later it takes a value that's passed as parameter.

I would much rather have the flag stay just a flag and make a new variable for the value, which can be `False` otherwise (seems the _create_response... function has this as default value to not update). So this `if` block could be replaced with something like:

```python
varname = delta.loc[ORBIT_DPP, DELTA] if update_deltap else False
```

omc3/model/accelerators/lhc.py
```diff
@@ -486,12 +488,48 @@ def _get_madx_main_sequence_loading(self) -> str:
                 "has to be specified (--lhcmode option missing?)."
             )
 
-    def get_update_correction_script(self, outpath: Path | str, corr_files: Sequence[Path | str]) -> str:
+    def get_update_correction_script(self, outpath: Path | str, corr_files: Sequence[Path | str], update_dpp: bool = False) -> str:
```
fsoubelet, line 491: Surely `update_dpp` here isn't expected to be a bool, considering the later checks and value passed to a function which expects `float | str`.

omc3/correction/handler.py
fsoubelet: Re: another comment on the `get_update_correction_script` later in my review: this should probably default as `None`, like the argument in that function since it should expect a bool (see comment there).

omc3/correction/handler.py
```diff
@@ -88,11 +94,15 @@ def correct(accel_inst: Accelerator, opt: DotDict) -> None:
...
-                resp_dict = response_twiss.create_response(accel_inst, opt.variable_categories, optics_params)
+                resp_dict = _update_response(
+                    accel_inst=accel_inst,
+                    corrected_model=corr_model,
+                    corrected_elements=corr_model_elements,
+                    delta=delta,
+                    optics_params=optics_params,
+                    variable_categories=opt.variable_categories,
+                    update_response=opt.update_response,
+                )
```
fsoubelet, line 105: Much better, thanks!

omc3/correction/handler.py
```diff
@@ -109,6 +119,60 @@ def correct(accel_inst: Accelerator, opt: DotDict) -> None:
...
+            LOG.info("Updating response via MAD-X.")
+            resp_dict = response_madx.create_fullresponse(accel_inst, variable_categories)
+        else:
+            LOG.info("Updating response via analytical formulae.")
+            resp_dict = response_twiss.create_response(accel_inst, variable_categories, optics_params)
+    
+    # reset model (not really necessary as only used here, but nicer)
+    accel_inst._model = original_model
+    accel_inst._elements = original_elements
+    return resp_dict
```
fsoubelet, line 173: Since we do not return anything to do with the "accelerator", wouldn't it be better to just make a copy of the `accel_inst` (i.e. via `copy.deepcopy`) and use that copy as argument in the function calls here? This way there is no worry about resetting the right attributes in the `accel_inst`.

Pingging @JoschD to get his opinion as well.

omc3/correction/handler.py
```diff
@@ -109,6 +119,60 @@ def correct(accel_inst: Accelerator, opt: DotDict) -> None:
...
+
+def _update_response(
+    accel_inst: Accelerator, 
+    corrected_model: pd.DataFrame,
+    corrected_elements: pd.DataFrame,
+    delta: pd.DataFrame, 
+    optics_params: Sequence[str], 
+    variable_categories: Sequence[str], 
+    update_response: bool | str,
+    ):
```
fsoubelet, line 131: Could you type hint the return?

omc3/correction/handler.py
fsoubelet: Otherwise the function here is quite nice, happy with it.

omc3/correction/handler.py
```diff
@@ -216,9 +280,9 @@ def _maybe_add_coupling_to_model(model: tfs.TfsDataFrame, keys: Sequence[str]) -
     return model
 
 
-def _create_corrected_model(twiss_out: Path | str, change_params: Sequence[Path], accel_inst: Accelerator) -> tfs.TfsDataFrame:
+def _create_corrected_model(twiss_out: Path | str, change_params: Sequence[Path], accel_inst: Accelerator, update_dpp: bool = False) -> tfs.TfsDataFrame:
```
fsoubelet, line 283: Same as below, surely `update_dpp` here isn't expected to be a bool, considering the later checks and value passed to a function which expects `float | str`.

omc3/model/accelerators/lhc.py
```diff
@@ -486,12 +488,48 @@ def _get_madx_main_sequence_loading(self) -> str:
...
-        for corr_file in corr_files:
+
+        # First set the dpp to the value in the accelerator model
+        madx_script += f"{ORBIT_DPP} = {self.dpp};\n"
+
+        for corr_file in corr_files:  # Load the corrections, can also update ORBIT_DPP
             madx_script += f"call, file = '{str(corr_file)}';\n"
-        madx_script += f"exec, do_twiss_elements(LHCB{self.beam}, '{str(outpath)}', {self.dpp});\n"
+        
+        if update_dpp: # If we are doing orbit correction, we need to ensure that a correct and matching is done (in get_update_deltap_script)
```
fsoubelet, line 500: Probably the "and" here is a typo?

omc3/model/accelerators/lhc.py
```diff
@@ -486,12 +488,48 @@ def _get_madx_main_sequence_loading(self) -> str:
...
+        )
+        return madx_script
+    
+    def get_tune_knobs(self) -> tuple[str, str]:
+        if self._uses_run3_macros():
+            return f"dQx.b{self.beam}_op", f"dQy.b{self.beam}_op"
+        elif self._uses_ats_knobs():
+            return f"dQx.b{self.beam}", f"dQy.b{self.beam}"
+        else:
+            return f"KQTD.B{self.beam}", f"KQTF.B{self.beam}"
```
fsoubelet, line 532: Nice.

omc3/model/accelerators/lhc.py
```diff
@@ -486,12 +488,48 @@ def _get_madx_main_sequence_loading(self) -> str:
...
+        
+        if update_dpp: # If we are doing orbit correction, we need to ensure that a correct and matching is done (in get_update_deltap_script)
+            madx_script += self.get_update_deltap_script(deltap=ORBIT_DPP)
+
+        madx_script += f'exec, do_twiss_elements(LHCB{self.beam}, "{str(outpath)}", {ORBIT_DPP});\n'
         return madx_script
+    
+    def get_update_deltap_script(self, deltap: float | str) -> str:
+        if not isinstance(deltap, str):
+            deltap = f"{deltap:.15e}"
```
fsoubelet, line 508: Maybe we can also check the `deltap` is some form of number and raise if not? I don't think we'd end up in this scenario but it could help diagnose since this is called quite deep into some code loops.

Not mandatory, feel free to ignore and resolve.

omc3/model/madx_macros/general.macros.madx
```diff
@@ -71,14 +71,14 @@ do_twiss_monitors(use_sequence, output_file, dpp): macro = {
 * preserve upper case characters.
 * @param dpp: delta p / p to use in the twiss command.
 */
-do_twiss_monitors_and_ips(use_sequence, output_file, dpp): macro = {
+do_twiss_monitors_and_ips(use_sequence, output_file, this_is_the_dpp): macro = {
```
fsoubelet, line 74: Not sure this is better. At least let's update the line above with the new parameter name.

tests/accuracy/test_global_correction.py
```diff
@@ -157,6 +160,54 @@ def test_lhc_global_correct(tmp_path, model_inj_beams, orientation):
...
+                optics_params = [f"{PHASE}X", f"{PHASE}Y"],
+                iterations=iteration,
+                update_response=update_response,
+                **model_inj_beams,
+            )
+            result = tfs.read(tmp_path / "changeparameters_iter.tfs", index=NAME)
+            
+            # Check if the output is correct within 5% (Beam 2 is not as accurate)
+            rtol = 5e-2 if iteration == 1 else 2e-2
+            assert np.isclose(dpp, -result[DELTA][ORBIT_DPP], rtol=rtol), f"Expected {dpp}, got {result[DELTA][ORBIT_DPP]}, diff: {dpp + result[DELTA][ORBIT_DPP]}, iteration: {iteration}"
```
fsoubelet, line 205: Which order of magnitude do we expect for the values here? The default of `atol` could be an issue.

fsoubelet (changes requested):
Few little things here and there, this is shaping up good.

tests/accuracy/test_global_correction.py
fsoubelet: Ah nevermind the values are given in the fixture and much higher than the `1e-08` of `atol`, my bad.

omc3/correction/handler.py
jgray-19: > I don't like this.
> 
> We define initially `update_deltap` as a bool which is used as a flag to check if we want to do this, but later it takes a value that's passed as parameter.
> 
> I would much rather have the flag stay just a flag and make a new variable for the value, which can be `False` otherwise (seems the _create_response... function has this as default value to not update). So this `if` block could be replaced with something like:
> 
> ```python
> varname = delta.loc[ORBIT_DPP, DELTA] if update_deltap else False
> ```

Ahh, sorry, didn't catch this, this is a merge artifact and should not have been pushed.

omc3/model/accelerators/lhc.py
jgray-19: It should be boolean, I should have never changed it to a float.

omc3/correction/handler.py
jgray-19: See above

omc3/model/accelerators/lhc.py
jgray-19: I meant it in a way that the "correct" and the "match" needs to be done

omc3/model/accelerators/lhc.py
jgray-19:
```suggestion
        if isinstance(deltap, str):
            deltap = f"{deltap:.15e}"
```

omc3/model/madx_macros/general.macros.madx
jgray-19: Done. I don't like it either, but it could affect anyone that includes `dpp` in their filename path otherwise.

omc3/model/accelerators/lhc.py
fsoubelet: Yes but maybe `isinstance(a, (int, float))` since an `int` would not trigger that check

omc3/model/madx_macros/general.macros.madx
fsoubelet: I had not thought of that. Good change then, let's update the parameter name and call it a day

omc3/model/accelerators/lhc.py
jgray-19: How about this [change](https://github.com/pylhc/omc3/pull/466/commits/63cd06cc546839fe0759e261faf9ba673ac938a6#diff-d1a4955c5b648d692ebecb63a82649e9614553aa5583a57647ee7995c81dc70fL500)

fsoubelet: Great, thanks

omc3/correction/handler.py
```diff
@@ -109,6 +116,49 @@ def correct(accel_inst: Accelerator, opt: DotDict) -> None:
...
+    Otherwise we go through the way of computing the response the user requested.
+
+    In case of dpp being modified in the correction, we have to add the new dpp value
+    to the accelerator instance, to run the `twiss` in madx at this dpp.
+    All other parameters are taken care of in the model/elements.
+    """
+    # update model by creating a copy of the accelerator instance
+    accel_inst_cp = copy.deepcopy(accel_inst)
+    accel_inst_cp.model = corrected_model
+    accel_inst_cp.elements = corrected_elements
```
JoschD, line 142: My issue with why I didn't do that, even though it is much nicer, is that the `elements` can be quite large and I was afraid that `deepcopy` will then double the memory footprint. We can leave it like that but we should be aware of this in case we ever see issues. On the other hand, we basically never update the response anyway and maybe this is all negligable ^^

jgray-19: Perhaps `copy.replace(obj, /, **changes)` would be better here? This would prevent elements being copied, as we're replacing it?

JoschD (commented):
Looks good to me, apart from the small changes we discussed in person. Thank you :)

jgray-19:
I've made some significant changes to get some of the code working. Let me know what you think, I'm not sure I like it all at moment...

jgray-19:
> > I've made some significant changes to get some of the code working. Let me know what you think, I'm not sure I like it all at moment...
> 
> Could it be an option to use the already existing modifier-functionality (https://github.com/pylhc/omc3/blob/594c5a9e5ec740e5797bfc89b14cdcf763ef45e4/omc3/model/accelerators/lhc.py#L402) on the accel_inst instead of passing the corr_files through the whole chain? Or would the corrections then be called at the wrong place?

It's possible, but it's difficult as the modifiers list is expected to be a list of files in the model directory. A possible solution could be to write these change parameter files to the model dir temporarily. Another solution could be to convert the files

jgray-19:
> > I've made some significant changes to get some of the code working. Let me know what you think, I'm not sure I like it all at moment...
> 
> Could it be an option to use the already existing modifier-functionality (
> 
> https://github.com/pylhc/omc3/blob/594c5a9e5ec740e5797bfc89b14cdcf763ef45e4/omc3/model/accelerators/lhc.py#L402
> ) on the accel_inst instead of passing the corr_files through the whole chain? Or would the corrections then be called at the wrong place?

See most recent commit

omc3/correction/handler.py
```diff
@@ -216,9 +267,9 @@ def _maybe_add_coupling_to_model(model: tfs.TfsDataFrame, keys: Sequence[str]) -
     return model
 
 
-def _create_corrected_model(twiss_out: Path | str, change_params: Sequence[Path], accel_inst: Accelerator) -> tfs.TfsDataFrame:
+def _create_corrected_model(twiss_out: Path | str, corr_files: Sequence[Path], accel_inst: Accelerator, update_dpp: bool = False) -> tfs.TfsDataFrame:
     """ Use the calculated deltas in changeparameters.madx to create a corrected model """
-    madx_script: str = accel_inst.get_update_correction_script(twiss_out, change_params)
+    madx_script: str = accel_inst.get_update_correction_script(twiss_out, corr_files, update_dpp)
```
JoschD, line 272: I like the renaming. We should maybe rename `changeparameters.madx` to something nicer in the future as well, but ... historical reasons ... (whoever came up with that name ... grrr)

omc3/correction/response_madx.py
```diff
@@ -83,31 +96,47 @@ def _generate_madx_jobs(
...
+    compute_deltap: bool = ORBIT_DPP in variables
+    no_dpp_vars = [var for var in variables if var != ORBIT_DPP]
+    vars_per_proc = int(np.ceil(len(no_dpp_vars) / num_proc))
 
     madx_job = _get_madx_job(accel_inst)
+    deltap_twiss = ""
+    # This is here only for multiple iteration of the global correction
+    # By including dpp here, it means that if deltap is in variables and dpp is not 0, the orbit and tune magnets change
+    # We have to be very careful that DELTAP_NAME is not used ANYWHERE else in MAD-X
+    if compute_deltap:
```
JoschD, line 112: Comment into if statement or at least a blank line above, so it is clear that it doesn't refer to any of the code above

omc3/correction/response_madx.py
```diff
@@ -83,31 +96,47 @@ def _generate_madx_jobs(
...
             var_idx = proc_idx * vars_per_proc + i
-            if var_idx >= len(variables):
+            if var_idx >= len(no_dpp_vars):
                 break
-            var = variables[var_idx]
+            var = no_dpp_vars[var_idx]
             incr_dict[var] = delta_k
             current_job += f"{var} = {var}{delta_k:+.15e};\n"
-            current_job += f"twiss, file='{str(temp_dir / f'twiss.{var}')}';\n"
+            current_job += f"twiss, file='{str(temp_dir / f'twiss.{var}')}'{deltap_twiss};\n"
```
JoschD, line 128: whoops! good catch

tests/unit/test_global_correction.py
```diff
@@ -1,15 +1,18 @@
...
+from omc3.correction.constants import VALUE, ORBIT_DPP
+from omc3.correction.handler import get_measurement_data, _update_response, _create_corrected_model
 from omc3.global_correction import OPTICS_PARAMS_CHOICES, CORRECTION_DEFAULTS, global_correction_entrypoint as global_correction
 from omc3.optics_measurements.constants import (BETA, DISPERSION, NORM_DISPERSION, F1001, F1010,
                                                 DISPERSION_NAME, BETA_NAME, PHASE_NAME,
                                                 NORM_DISP_NAME, F1010_NAME, TUNE, PHASE,
-                                                F1001_NAME)
+                                                F1001_NAME, EXT, DELTA)
 from omc3.scripts.fake_measurement_from_model import generate as fake_measurement
 from tests.accuracy.test_global_correction import get_skew_params, get_normal_params
```
JoschD, line 17: `pd` and `ORBIT_DPP` do not seem to be needed. 
This could also use an auto-format. (ruff, isort, black, etc.)

tests/unit/test_global_correction.py
```diff
@@ -106,3 +109,84 @@ def test_lhc_global_correct_methods(tmp_path, model_inj_beams, method):
 
     if method == "pinv":
         assert len(correction.index) > n_correctors  # unless by accident
+
+@pytest.mark.basic
+def test_update_response(tmp_path, model_inj_beams):
+    """ Tests if the response is updated. """
+    # create the accelerator instance
+    beam = model_inj_beams.beam
```
JoschD, line 117: not used

tests/unit/test_global_correction.py
```diff
@@ -106,3 +109,84 @@ def test_lhc_global_correct_methods(tmp_path, model_inj_beams, method):
 
     if method == "pinv":
         assert len(correction.index) > n_correctors  # unless by accident
+
+@pytest.mark.basic
+def test_update_response(tmp_path, model_inj_beams):
```
JoschD, line 114: 🙏🏻🙏🏻

tests/unit/test_global_correction.py
```diff
@@ -106,3 +109,84 @@ def test_lhc_global_correct_methods(tmp_path, model_inj_beams, method):
...
+        corrected_elements=corr_model,
+        optics_params=optics_params,
+        corr_files=[],
+        variable_categories=[knob],
+        update_dpp=False,
+        update_response=True
+    )
+    for key in new_resp_dict.keys(): # The new dict will only have the keys for the selected optics_params
+        # The values are not the same, as the model has changed.
+        assert not ref_resp_dict[key].equals(new_resp_dict[key])
```
JoschD, line 187: I think this is not a good test, as ref_resp was created with madx. You should create a ref also with the twiss repsonse and original model first.

tests/unit/test_global_correction.py
```diff
@@ -106,3 +109,84 @@ def test_lhc_global_correct_methods(tmp_path, model_inj_beams, method):
...
+        assert_frame_equal(ref_resp_dict[key], new_resp_dict[key])
+
+    corr_file = tmp_path / "corr.madx"
+    with open(corr_file, "w") as f:
+        f.write(f"{knob} = {knob}{delta.loc[knob, DELTA]:+.16e};")
+
+    corr_model = _create_corrected_model(corr_model_path, [corr_file], accel_inst, False)
+    assert not same_model.equals(corr_model)
+
+    # Now for response_madx with the model changed
```
JoschD, line 155: Actually, that the model has changed is irrelevant for the madx-response update. 

I think you should create the updated model only after the madx-response  test, as it belongs to the twiss-response update, which uses the data from the twiss (i.e. the "df_elements") to calcualte the response.

JoschD (changes requested):

tests/unit/test_global_correction.py
jgray-19: I installed ruff! Hope you like it :)

JoschD (dismissed):
To me it looks tip top

JoschD (approved):

PR #489, opened 2025-03-10
SPS Model Creation

JoschD:
This implements SPS model creation.

- I implemented the full range of model creators (without Best-Knowledge), i.e. Nominal, Segment and Correction.
- The tests for SPS are now all in one single test file, as I feel this also makes a lot of sense to sort things that way and will help also future accelerators to implement similar tests.

Sorry, that there are so many changed files (again). 
 - A lot of them are again test-input files.
  I also had to update some of the LHC test files, which were created with the 2025 model, as the lhc-sequence for 2025 was updated on monday, leading to some test failures (e.g. a missing BPM)
- I updated/simplified some more things with the model creation, which should allow easy implementing new accelerators in the future (see the whole sps_model_creator file). These were only minor changes, but affected some files. 
- Same for additional constants I introduced.


omc3/model/accelerators/sps/correctors.json
```diff
@@ -1,226 +1,226 @@
...
-        "kdva61932",
-        "kdv62107",
-        "kdv62307",
-        "kdv62507",
-        "kdv62707",
-        "kdv62907",
-        "kdv63107",
-        "kdv63307",
-        "kdv63507"
+        "kmdh10207",
```
fsoubelet, line 3: Trusting you on these ones. Can you just let me know where you got the updated info from? Directly lattice files?

omc3/model/model_creators/abstract_model_creator.py
```diff
@@ -264,32 +260,13 @@ def prepare_symlink(self):
...
-        msg = f"Couldn't find modifier {modifier}.\nAlso tried in {accel.model_dir}"
-        if accel.acc_model_path is not None:
-            msg += f" and in {accel.acc_model_path}"
-        raise FileNotFoundError(msg)
-
+    def prepare_modifiers(self):
+        """ Loop over the modifiers and make them full paths if found. """
+        accel: Accelerator = self.accel
+        if accel.modifiers is not None:
+            accel.modifiers = [accel.find_modifier(m) for m in accel.modifiers]
```
fsoubelet, line 268: Much nicer

omc3/model/model_creators/sps_model_creator.py
```diff
@@ -42,21 +44,42 @@ def check_accelerator_instance(self) -> None:
...
                 "The accelerator definition is incomplete: "
-                "SPS model creation only works with acc-models, but `acc-model-path` was not set. "
+                "Model directory (outputdir option) was not given."
+            )
+
+        if accel.modifiers is None or not len(accel.modifiers):
+            raise AcceleratorDefinitionError(
+                "The accelerator definition is incomplete: No modifiers could be found. "
+                "Hint: If the accelerator class was instantiated from a model dir, "
+                f"make sure {JOB_MODEL_MADX_NOMINAL} exists and contains the '{MODIFIER_TAG}' tag."
```
fsoubelet, line 64: For me to be sure: we also consider something like the `opticsfile` / `strength file` to be a modifier right? This way it makes sense to have that error

omc3/response_creator.py
```diff
@@ -125,6 +125,8 @@ def create_response_entrypoint(opt: DotDict, other_opt) -> dict[str, pd.DataFram
     """Entry point for creating pandas-based response matrices.
 
     The response matrices can be either created by response_madx or TwissResponse.
+
+    TODO: Add supprt to give fetcher? But for know we assume the acc-model path is still in the model_dir.
```
fsoubelet, line 129: support

tests/unit/test_sps.py
```diff
@@ -0,0 +1,299 @@
...
+    _check_linear_optics_files,
+    _check_nonlinear_optics_files,
+)
+from tests.unit.test_model_creator import check_accel_from_dir_vs_options
+
+SPS_DIR = INPUTS / "sps_data"
+SPS_MODEL_DIR = SPS_DIR / "model_Q20_noacd"
+Q20_STRENGTHS_FILE = "lhc_q20.str"
+
+class TestModelCrationSPS:
```
fsoubelet, line 65: creation

tests/unit/test_sps.py
```diff
@@ -0,0 +1,299 @@
...
+            fetch=Fetcher.PATH,
+            path=acc_models_sps_2025,
+            modifiers=[Q20_STRENGTHS_FILE],
+        )
+        accel = create_instance_and_model(
+            outputdir=tmp_path, type=CreatorType.NOMINAL, logfile=tmp_path / "madx_log.txt", **accel_opt
+        )
+
+        if not use_acdipole:
+            accel.drv_tunes = None  # cannot read that form model-dir when TWISS-AC is not created
```
fsoubelet, line 85: from

fsoubelet (commented):
Cool. Very minor comments here and there.

omc3/model/accelerators/sps/correctors.json
JoschD: "Updated" is wrong. this file is new. The first info was more a guess, based on LHC nomenklature, the change was because I checked the sequence.

omc3/model/model_creators/sps_model_creator.py
JoschD: Yes. Basically everything that changes the parameters, once the sequence is built, we consider as modifiers. At least this is how it was in the LHC, where the optics files were always "modifiers". It also makes it easier from the gui to not having to deal with different flags for different machines. 
Sadly, the implementation of PS and PSB does not follow this philosophy.

omc3/model/accelerators/sps/correctors.json
fsoubelet: ok, thanks

omc3/model/model_creators/sps_model_creator.py
fsoubelet: Got it, definitely agree it is better this way.

PR #490, opened 2025-03-14
SbS Propagable Refactor

JoschD:
Splits the Propagables in SbS into different files in preparation to add more propagables. It was getting a bit cluttered.


omc3/segment_by_segment/propagables/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+
+This module exposes the main classes for propagables, to avoid making the 
+imports so long and complicated.
+"""
+from omc3.segment_by_segment.propagables.abstract import Propagable  # probably only used for type-hinting
+from omc3.segment_by_segment.propagables.utils import PropagableBoundaryConditions   #noqa  -> expose
+from omc3.segment_by_segment.propagables.utils import PropagableColumns   #noqa  -> expose
+from omc3.segment_by_segment.propagables.phase import Phase
+from omc3.segment_by_segment.propagables.beta import BetaPhase
+from omc3.segment_by_segment.propagables.alpha import AlphaPhase
```
fsoubelet, line 18: We should make an `__all__` with things we want to publicly expose.

omc3/segment_by_segment/propagables/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+
+def get_all_propagables() -> tuple[Propagable, ...]:
+    """ Return all defined Propagables. """
+    return Phase, BetaPhase, AlphaPhase
```
fsoubelet, line 22: Feels a bit weird to have a function for three hardcoded things, might as well be a tuple directly.

omc3/segment_by_segment/propagables/abstract.py
```diff
@@ -0,0 +1,152 @@
...
+
+from abc import ABC, abstractmethod
+from functools import cache
+from typing import TYPE_CHECKING
+
+import pandas as pd
+
+from omc3.definitions.constants import PLANES
+from omc3.definitions.optics import OpticsMeasurement
+from omc3.segment_by_segment import propagables  # don't import alpha/beta etc directly ! cyclic imports !
```
fsoubelet, line 17: Oof, big smell that we are potentially doing cyclic imports...

omc3/segment_by_segment/propagables/abstract.py
```diff
@@ -0,0 +1,152 @@
...
+        """Deviations between backward propagated models with and without correction."""
+        ...
+    
+    @abstractmethod
+    def add_differences(self, segment_diffs: SegmentDiffs):
+        """This function calculates the differences between the propagated 
+        forward and backward models and the measured values.
+        It then adds the results to the segment_diffs class 
+        (which writes them out, if its ``allow_write`` is set to ``True``)."""
+        ...
```
fsoubelet, line 152: eol at eof please, or just auto formatting on this file

fsoubelet (approved):
I know it's draft, but I stay ahead of the curve on this one. It's a refactor and as far as I can tell there is no change in the moved code.

Minor comments as always, feel free to ignore.
I'd be happy to merge this PR and have another one implementing coupling SbS from the new structure.

omc3/segment_by_segment/propagables/__init__.py
fsoubelet: And rather than updating a `return` statement we would just update the tuple?

omc3/sbs_propagation.py
```diff
@@ -139,6 +138,7 @@ def segment_by_segment(opt: DotDict, accel_opt: dict | list) -> dict[str, Segmen
 
     accel: Accelerator = model_manager.get_accelerator(accel_opt)
     accel.model_dir = _check_output_directory(opt.output_dir, accel)
+    save_config(accel.model_dir, opt, __file__, accel_opt)
```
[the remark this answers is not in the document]
JoschD: No, the timestamp is added to the config file, as always

omc3/definitions/formats.py
```diff
@@ -15,7 +15,7 @@
 def get_config_filename(script):
     """Default Filename for config-files. Call from script with ``__file__``."""
     return CONFIG_FILENAME.format(
-        script=Path(script).name.split('.')[0],
+        script=Path(script).stem,
```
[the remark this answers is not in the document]
JoschD: I think it was me, though. Sorry... ?

omc3/sbs_propagation.py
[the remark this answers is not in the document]
fsoubelet: Yeah I should have remembered that, my bad.

omc3/definitions/formats.py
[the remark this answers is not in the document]
fsoubelet: what friday evening PRs do to a man

omc3/sbs_propagation.py
```diff
@@ -433,6 +433,8 @@ def _copy_files_from_model_dir(model_dir: Path, output_dir: Path) -> None:
 def _bpm_is_in_meas(bpm_name: str, meas: OpticsMeasurement) -> bool:
     """ Check if the bpm_name is in the measurement in both planes.
     Possible improvement: Check if the error is too high?
+
+    TODO: This does not work with single-plane BPM machines, such as SPS. What to do?
```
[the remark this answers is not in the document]
JoschD: Without beta_phase the whole sbs will not work, so I think it's fine to just crash here. 

Adaption wise: Not so easy. You would really have to run things twice and do a propagation per plane. Maybe there is some easier way, that I am not seeing, but on the other hand we do not have to think about this for now, as we do not really need it in SPS. Maybe I should remove even the possibility, though (i.e. remove the segment creator)

omc3/segment_by_segment/propagables/__init__.py
JoschD: Is this not only encouraging people to do `from propagables import *` ?
Also `__all__` here would just repeat all the things inside (do not want to hide anything)

omc3/segment_by_segment/propagables/abstract.py
JoschD: As this problem only comes here in the Abstract class and I should not need to touch this code again, I think we will be fine. Also you will notice immediately with the tests

omc3/segment_by_segment/propagables/__init__.py
JoschD: Not that it makes much difference, but as I was also thinking the same I will do that ^^

omc3/segment_by_segment/propagables/__init__.py
fsoubelet: No, it does not encourage, it is a safeguard in case someone does. I believe IDEs also try to inspect __all__ if possible when suggesting completions. It's better practice

PR #497, opened 2025-03-28
Bad BPMs Summary: NaNs

JoschD:
Added the NaNs (see https://github.com/pylhc/omc3/tree/0.21.0 ) as a known filter/reason to the summary.


fsoubelet (approved):
A Josch PR with less than 50 modified files? Must be my birthday!

PR #502, opened 2025-04-04
Ruff configuration

fsoubelet:
As discussed, we want to converge on a simple setup for formatting + linting and I offer a configuration using `Ruff`. It's fast, does formatting + linting (+ a [type checker](https://x.com/charliermarsh/status/1884651482009477368) at some point), easy to install as a standalone etc.

Ruff, like Black, comes quite opinionated (actually, [they're compatible](https://docs.astral.sh/ruff/faq/#is-the-ruff-linter-compatible-with-black)) and I suggest minimal changes to the default settings.

**Changes in this PR:**

- Configuration in its own table in `pyproject.toml`. Very open to discussion.
- A GA file to run `Ruff` linting on commits in pull requests.
- Small fixes for the current linting complaints.

**Suggested philosophy:**

None of this runs automated. I suggest the following:

- Enforcing a soft-rule for formatting on new code and when refactoring. We should eventually converge to a coherent codebase.
- Relatively nice tolerance to the use of `# fmt: off`, `# fmt: on` and `# noqa: RULE` when it makes sense (preserving mathematical formatting, accepting "no-go" varnames where justified etc).
- Leaving the GA workflow as a pass-fail test (its default). Currently it passes, so only PR modifications & new code could break it, and we would be aware at review without the need to checkout the PR branch, run locally etc.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -18,3 +18,4 @@ jobs:
       - uses: astral-sh/ruff-action@v3
         with:
           version: "latest"
+          src: "omc3"
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,0 +1,21 @@
+name: Ruff linting
+
+defaults:
+  run:
+    shell: bash
+
+# Runs on any push event in a PR
+on:
+  pull_request:
+  push:
+
+
+jobs:
+  linter:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      # This runs ruff check by default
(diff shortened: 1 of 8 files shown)


.github/workflows/checks.yml
```diff
@@ -0,0 +1,21 @@
+name: Ruff linting
+
+defaults:
+  run:
+    shell: bash
+
+# Runs on any push event in a PR
+on: [ push, pull_request ]
```
jgray-19: I think technically, it will run on any push to the repository and any pull request event

JoschD (approved):

PR #507, opened 2025-04-27
Kmod lumi imablance: IP Naming

### LINKED ISSUE #506: [Bug]: K-mod lumi imbalance overwrites file
### Bug Description

When the kmod lumi imbalance is run with multiple IPs, there are some issues

a) the file name is the same for all imbalances, so it overwrites
b) In the .ini file also not-given IPs are declared as "given as DF"
c) When run as a script alone, kmod_lumi_imbalance does not initialize the logger (it seems) or if it does, there should be at least some logging done

JoschD:
Fixes #506

### DIFF
--- a/omc3/scripts/kmod_lumi_imbalance.py
+++ b/omc3/scripts/kmod_lumi_imbalance.py
@@ -113,7 +113,7 @@ def calculate_lumi_imbalance(opt: DotDict) -> tfs.TfsDataFrame:
         output_path.mkdir(parents=True, exist_ok=True)
         opt_cp = copy.copy(opt)
         for ip in IPS:
-            if not isinstance(opt_cp[ip], (Path, str)):
+            if opt_cp[ip] is not None and not isinstance(opt_cp[ip], (Path, str)):
                 opt_cp[ip] = "(was provided as DataFrame)"
         save_config(output_path, opt_cp, __file__)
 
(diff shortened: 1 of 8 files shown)


omc3/scripts/kmod_lumi_imbalance.py
```diff
@@ -187,11 +190,13 @@ def get_lumi_imbalance_df(**kwargs) -> tfs.TfsDataFrame:
     Args:
         ipA (tfs.TfsDataFrame): a `TfsDataFrame` with the averaged results from a kmod analysis, for IP_A.
         ipB (tfs.TfsDataFrame): a `TfsDataFrame` with the averaged results from a kmod analysis, for IP_B.
-        (Actually, any name that ends with an integer is fine.)
+        (Actually, any name is fine.)
```
fsoubelet, line 193: You mean for the provided args, right?

fsoubelet: Feels like lines 131-132 expect some form of nomenclature

fsoubelet (approved):

omc3/scripts/kmod_lumi_imbalance.py
JoschD: I want to say that any key for the `kwargs` is fine, just by convention it should be something like "ipA" "ipB" but it doesn't really matter for the function itself. I think the part with the "ends with an integer" was a remainder of some old code, where I only used the  IP number. As I do kind of now in l.132/l.133, for the "ip" labels in the name. But that is a different function and also does not need to be an integer. IPH or IPX would work as well, I am only hoping that the last letter of the name is the main identifier ^^

PR #510, opened 2025-05-09
KEK-specific bugfixes: Model energy definition and BPM naming convention

wvangoet:


omc3/model/accelerators/accelerator.py
```diff
@@ -185,7 +185,10 @@ def init_from_model_dir(self, model_dir: Path) -> None:
             bpm_mask = self.elements.index.str.match(self.RE_DICT[AccElementTypes.BPMS])
             self.model = self.elements.loc[bpm_mask, :]
         self.nat_tunes = [float(self.model.headers["Q1"]), float(self.model.headers["Q2"])]
-        self.energy = float(self.model.headers["ENERGY"])  # always 450GeV because we do not set it anywhere properly...
+        try:
+            self.energy = float(self.model.headers["ENERGY"])  # always 450GeV because we do not set it anywhere properly...
+        except KeyError: #KEK model does not have energy in the header
+            pass
```
fsoubelet, line 191: I would prefer that we change this line to
```python
self.energy = float(self.model.headers.get("ENERGY", 0))
```

or with some other default value (maybe 450GeV, willing to get @JoschD's input here).

This way we can try simply have a check that `self.energy` is set in the actual accelerator class, and not fear a missing attribute.

omc3/model/accelerators/skekb.py
```diff
@@ -75,6 +75,11 @@ class SKekB(Accelerator):
     """KEK's SuperKEKB accelerator."""
     NAME = "skekb"
     RINGS = ("ler", "her")
+    RE_DICT: dict[str, str] = {
+        AccElementTypes.BPMS: r"^M*",
+        AccElementTypes.MAGNETS: r".*",  # has a variety of names...
+        AccElementTypes.ARC_BPMS: r"^M*",
+    } # naming convention for KEK elements
```
fsoubelet, line 82: Nice. Seems like `AccElementTypes` is not imported though.

fsoubelet (changes requested):
Thanks! Some suggestions.

I would like to have a quick test on this as well before we merge.

fsoubelet:
Also @wvangoet, since you are already part of pylhc/omc-team it would be better if you clone this repository directly and branch from that, so the CI can run without complaining about tokens state.

omc3/model/accelerators/accelerator.py
JoschD: The 40GeV comment refers to the fact, that we are not necessarily setting it properly in the LHC model creation - or at least had not been setting in in the past, when I wrote this comment. So the default of 450 was only because that's the default set in madx for lhc... I think we fixed this now, though, for LHC. 

For other machines: I'd prefer `None` over `0` because that's a clearer indication that the energy has not been set (e.g. if someone later has `if energy < 6800` somewhere I'd prefer that to fail with an error, not to silently pass because someone forgot to set the energy.

JoschD: and I was thinking about using `get` here, but then you will have issues with the `float`, so I wrote the try-except when suggesting this fix to Wietse.

But yeah, also in favour of adding a test (maybe extended) to see if `None` makes the code crash down the line.

fsoubelet: I'd rather have a `.get` with a `None` default value than risk getting an attribute or just... the attribute not existing.
I don't think the cast to float is that important, as Python upcasts whenever it has the chance anyway.

I'd prefer we be clear on the state of an `Accelerator`'s instance and handle the potential `None` default later :)

JoschD: I'm not sure if I understand your point. 
`self.energy = None` is set above, so the try-except line is basically the same as `.get(XXX, None)` with the additional benefit of casting to float, if an energy is present

fsoubelet: Eh, my bad. `self.energy = self.model.headers.get("ENERGY", None)` is perfectly fine then!

JoschD: You mean try-except? I am trying to argue for try-except here xD

fsoubelet: I like the conciseness. Do we care to cast to float?

wvangoet: For KEK data, I am able to run through the optics analysis while energy = None but I'm only really looking at RDT data so other functionalities might still be broken. Let me know when it's decided and I'll make a new pull request.

omc3/model/accelerators/skekb.py
JoschD: @wvangoet This is the more important issue to fix, imho. Not sure how this can even work for you

PR #512, opened 2025-05-13
Handle TbtData objects as input files

Mael-Le-Garrec:
Now that XSuite exists, it's possible to run tracking simulations directly in python. Using harpy requires to create a TbtData object and write an SDDS.  
This PR skips the SDDS writing step, as now `TbtData` objects can be loaded directly.

Harpy's logic was to use the filename to determine the output, a new argument `tbt_name` is introduced for the user to choose the name of the output.

Linked to https://github.com/pylhc/turn_by_turn/pull/22

### DIFF
--- a/omc3/harpy/handler.py
+++ b/omc3/harpy/handler.py
@@ -39,7 +39,11 @@ def run_per_bunch(tbt_data, harpy_input):
     """
     model = None if harpy_input.model is None else tfs.read(harpy_input.model, index=COL_NAME).loc[:, 'S']
     bpm_datas, usvs, lins, bad_bpms = {}, {}, {}, {}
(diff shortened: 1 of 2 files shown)


JoschD:
I was thinking that we could possibly add an optional `file` or `info` or `name` attribute to the `TbTData` class (that we might as well even set automatically in the reader, but not neccessay for this to work here). And then we could skip the `tbt_name` argument here and say "try this attribute, if this is not set, use str(files) as before". That way you can just set the name manually directly at the object, keep better track of it and we do not have an additional argument to hio. But that's just an idea, might have downsides that I don't see at the moment.

JoschD:
@fsoubelet @jgray-19 what do you think? about my last comment?

fsoubelet:
> @fsoubelet @jgray-19 what do you think? about my last comment?

The bypass would be in hole-in-one?

PR #519, opened 2025-06-26
Tightening the screws, part 1

fsoubelet:
Now that we have a linting check in CI, I am suggesting adding some more rules to the `Ruff` setup on top of the default ones.

This PR introduces a list of rules, some of which I've activated and fixed (hence the file changes). All other changes in said files are the removal of trailing spaces that my IDE did.

~~There was one non-straightforward case of `NPY002` in `omc3/scripts/fake_measurement_from_model.py`, where a `np.random.seed` was set without random number generation, most likely to propagate to all called functions.~~

~~Considering the [numpy docs](https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html) actively discourage doing this, and recommend a per-generator seeding I have propagated the seed down to called functions and let those use it when creating their own Generators.~~

@JoschD @jgray-19 I am eager to get your opinion on the list of rules I have envisioned, in the `[ruff.lint]` table of the `pyproject.toml` file, including those I have not yet activated. You can look for their meaning and check at https://docs.astral.sh/ruff/rules/. They can come in a later PR.

### DIFF
--- a/omc3/correction/response_io.py
+++ b/omc3/correction/response_io.py
@@ -106,14 +106,14 @@ def write_varmap(path: Path, varmap: dict[str, dict[str, pd.Series]]):
 # Helper -----------------------------------------------------------------------
 
 
-def _check_keys(store: pd.HDFStore, keys: Sequence[str], id: str):
+def _check_keys(store: pd.HDFStore, keys: Sequence[str], file_id: str):
     if keys is None:
         return
 
     groups = _main_store_groups(store)
     not_found = [k for k in keys if k not in groups]
     if len(not_found):
-        raise ValueError(f"The following keys could not be found in {id} file:" f" {', '.join(not_found)}")
+        raise ValueError(f"The following keys could not be found in {file_id} file:" f" {', '.join(not_found)}")
 
 
 def _main_store_groups(store: pd.HDFStore) -> set[str]:
--- a/omc3/plotting/utils/colors.py
+++ b/omc3/plotting/utils/colors.py
@@ -7,7 +7,7 @@ Helper functions to handle colors in plots.
 import colorsys
 from itertools import cycle
 
-import matplotlib
+import matplotlib as mpl
 from matplotlib import colors as mc
 
 
@@ -76,5 +76,5 @@ def change_ebar_alpha_for_line(ebar, alpha):
 def change_ebar_alpha_for_axes(ax, alpha):
     """Wrapper for change_ebar_alpha_for_line to change all in one axes."""
     for ebar in ax.containers:
-        if isinstance(ebar, matplotlib.container.ErrorbarContainer):
+        if isinstance(ebar, mpl.container.ErrorbarContainer):
             change_ebar_alpha_for_line(ebar, alpha)
--- a/omc3/plotting/utils/colors.py
+++ b/omc3/plotting/utils/colors.py
@@ -7,8 +7,7 @@ Helper functions to handle colors in plots.
 import colorsys
 from itertools import cycle
 
-from matplotlib import colors as mc
-from matplotlib.container import ErrorbarContainer
+import matplotlib as mpl
 
 
 def get_mpl_color(idx=None):
@@ -52,14 +51,11 @@ def change_color_brightness(color, amount=0.5):
         raise ValueError("The brightness change has to be between 0 and 2."
                          " Instead it was {}".format(amount))
     try:
-        c = mc.cnames[color]
+        c = mpl.colors.cnames[color]
     except KeyError:
         c = color
 
-    try:
-        c = colorsys.rgb_to_hls(*mc.ColorConverter().to_rgb(c))  # matplotlib 1.5
-    except AttributeError:
-        c = colorsys.rgb_to_hls(*mc.to_rgb(c))  # matplotlib > 2
+    c = colorsys.rgb_to_hls(*mpl.colors.to_rgb(c))
     return colorsys.hls_to_rgb(c[0], max(0, min(1, amount * c[1])), c[2])
 
 
@@ -76,5 +72,5 @@ def change_ebar_alpha_for_line(ebar, alpha):
 def change_ebar_alpha_for_axes(ax, alpha):
     """Wrapper for change_ebar_alpha_for_line to change all in one axes."""
     for ebar in ax.containers:
-        if isinstance(ebar, ErrorbarContainer):
+        if isinstance(ebar, mpl.container.ErrorbarContainer):
             change_ebar_alpha_for_line(ebar, alpha)
--- a/omc3/plotting/utils/colors.py
+++ b/omc3/plotting/utils/colors.py
@@ -7,8 +7,8 @@ Helper functions to handle colors in plots.
 import colorsys
 from itertools import cycle
 
-import matplotlib as mpl
 from matplotlib import colors as mc
+from matplotlib.container import ErrorbarContainer
 
 
 def get_mpl_color(idx=None):
@@ -76,5 +76,5 @@ def change_ebar_alpha_for_line(ebar, alpha):
 def change_ebar_alpha_for_axes(ax, alpha):
     """Wrapper for change_ebar_alpha_for_line to change all in one axes."""
     for ebar in ax.containers:
-        if isinstance(ebar, mpl.container.ErrorbarContainer):
+        if isinstance(ebar, ErrorbarContainer):
             change_ebar_alpha_for_line(ebar, alpha)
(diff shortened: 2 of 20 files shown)


fsoubelet:
For some reason the `Ruff` CI has not failed on this check 🤔

omc3/correction/response_io.py
```diff
@@ -106,14 +106,14 @@ def write_varmap(path: Path, varmap: dict[str, dict[str, pd.Series]]):
 # Helper -----------------------------------------------------------------------
 
 
-def _check_keys(store: pd.HDFStore, keys: Sequence[str], id: str):
+def _check_keys(store: pd.HDFStore, keys: Sequence[str], file_id: str):
```
fsoubelet, line 109: Here `id` shadowed a [builtin function](https://docs.python.org/3/library/functions.html#id)

omc3/scripts/fake_measurement_from_model.py
```diff
@@ -202,24 +203,27 @@ def generate(opt) -> dict[str, tfs.TfsDataFrame]:
     global_correction_entrypoint uses (same format you would get from hole_in_one).
     """
     LOG.info("Generating fake measurements.")
+
     # prepare data
-    np.random.seed(opt.seed)
```
fsoubelet, line 206: This was removed as global random state is actively discouraged by `numpy`

omc3/scripts/fake_measurement_from_model.py
```diff
@@ -251,18 +255,19 @@ def generate(opt) -> dict[str, tfs.TfsDataFrame]:
 # Main Creator Functions -------------------------------------------------------
 
 def create_beta(df_twiss: pd.DataFrame, df_model: pd.DataFrame, parameter: str,
-                relative_error: float, randomize: Sequence[str], headers: dict):
+                relative_error: float, randomize: Sequence[str], headers: dict,
+                seed: int | None = None):
```
fsoubelet, line 259: This now takes a `seed` and passes it down.

omc3/scripts/fake_measurement_from_model.py
```diff
@@ -274,32 +279,35 @@ def create_beta(df_twiss: pd.DataFrame, df_model: pd.DataFrame, parameter: str,
 
 
 def create_dispersion(df_twiss: pd.DataFrame, df_model: pd.DataFrame, parameter: str,
-                      relative_error: float, randomize: Sequence[str], headers: dict):
+                      relative_error: float, randomize: Sequence[str], headers: dict,
+                      seed: int | None = None):
```
fsoubelet, line 283: This now takes a `seed` and passes it down.

omc3/scripts/fake_measurement_from_model.py
```diff
@@ -274,32 +279,35 @@ def create_beta(df_twiss: pd.DataFrame, df_model: pd.DataFrame, parameter: str,
...
     df = append_model_param(df, df_model, parameter)
     df = append_model_s_and_phaseadv(df, df_model, planes=plane)
     df.headers = headers.copy()
     return {f'{DISPERSION_NAME}{plane.lower()}': df}
 
 
 def create_phase(df_twiss: pd.DataFrame, df_model: pd.DataFrame, parameter: str,
-                 relative_error: float, randomize: Sequence[str], headers: dict):
+                 relative_error: float, randomize: Sequence[str], headers: dict,
+                 seed: int | None = None):
```
fsoubelet, line 296: This now takes a `seed` and passes it down.

omc3/scripts/fake_measurement_from_model.py
```diff
@@ -274,32 +279,35 @@ def create_beta(df_twiss: pd.DataFrame, df_model: pd.DataFrame, parameter: str,
...
-    dict_tot = create_total_phase(df_twiss, df_model, parameter, relative_error, randomize, headers)
+    dict_tot = create_total_phase(df_twiss, df_model, parameter, relative_error, randomize, headers, seed)
     results.update(dict_tot)
     return results
 
 
 def create_phase_advance(df_twiss: pd.DataFrame, df_model: pd.DataFrame, parameter: str,
-                         relative_error: float, randomize: Sequence[str], headers: dict):
+                         relative_error: float, randomize: Sequence[str], headers: dict,
+                         seed: int | None = None):
```
fsoubelet, line 310: This now takes a `seed` and uses it in for its own `Generator`. It passes to `_get_random_errors` if errors are to be randomized too.

omc3/scripts/fake_measurement_from_model.py
```diff
@@ -538,17 +551,19 @@ def _get_loop_parameters(parameters: Sequence[str], errors: Sequence[float] | No
     return zip(parameters, errors)
 
 
-def _get_random_errors(errors: np.array, values: np.array) -> np.array:
+def _get_random_errors(errors: np.array, values: np.array, seed: int | None = None) -> np.array:
```
fsoubelet, line 554: This now takes a `seed` and uses it in for its own `Generator`.

omc3/utils/parsertools.py
```diff
@@ -38,9 +38,9 @@ def require_param(name: str, parameters: dict, options: dict):
         raise AttributeError(f"Missing flag `{name}`.\nUsage:\n{_get_help_str(name, parameters)}")
 
 
-def _maybe_get(space: str, category: str, map: dict)  -> str:
-    if category in map:
-        return f"{space}{category}: {map[category]}\n"
+def _maybe_get(space: str, category: str, mapping: dict)  -> str:
```
fsoubelet, line 41: Here `map` shadowed a [builtin function](https://docs.python.org/3/library/functions.html#map)

pyproject.toml
```diff
@@ -164,11 +164,32 @@ indent-width = 4
 # Allow unused variables when underscore-prefixed.
 dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 ignore = [
+  "E501",  # line too long
```
fsoubelet, line 167: This should be temporary, at least until we've done a refactoring pass for formatting. But it's a big refactor commit and not for today.

pyproject.toml
```diff
@@ -164,11 +164,32 @@ indent-width = 4
...
+    # "ISC",      # Good use of string concatenation
+    "BLE",      # disallow catch-all exceptions
+    "ICN",      # Use common import conventions
+    # "RET",      # Good return practices
+    # "SIM",      # Common simplification rules
+    # "TID",      # Some good import practices
+    # "TC",       # Enforce importing certain types in a TYPE_CHECKING block
+    # "PTH",      # Use pathlib instead of os.path
+    "NPY",      # Some numpy-specific things
+]
```
fsoubelet, line 192: Some more here are quite straight-forward but I wanted to avoid a gigantic PR.

fsoubelet:
Well, those tests passed locally :(

pyproject.toml
```diff
@@ -164,11 +164,32 @@ indent-width = 4
...
+    "FA",       # Enforce from __future__ import annotations
+    "FBT",      # detect boolean traps
+    # "ISC",      # Good use of string concatenation
+    "BLE",      # disallow catch-all exceptions
+    "ICN",      # Use common import conventions
+    # "RET",      # Good return practices
+    # "SIM",      # Common simplification rules
+    # "TID",      # Some good import practices
+    # "TC",       # Enforce importing certain types in a TYPE_CHECKING block
+    # "PTH",      # Use pathlib instead of os.path
```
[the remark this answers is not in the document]
fsoubelet: Had a quick look and this one shouldn't be the most work to pass. Quite a few errors but `Ruff` even tells you what the replacement is.

fsoubelet:
On further investigation, I see that we have terrible use of `numpy.random` in the tests themselves. There are many instances of `numpy.random.seed(...)`, which *each* modify the global state of the program.

So this can move on, I've commented out the rule and reverted the changes to `fake_measurement_from_model.py`, to be left to its own dedicated branch and PR. That one will be ugly.

omc3/plotting/utils/colors.py
```diff
@@ -7,8 +7,8 @@
 import colorsys
 from itertools import cycle
 
-import matplotlib
 from matplotlib import colors as mc
+from matplotlib.container import ErrorbarContainer
```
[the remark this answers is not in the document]
jgray-19: I think I would prefer only to do `import matplotlib as mpl` when `mpl` is accessed for a few things, else, just import what we need.

jgray-19 (approved):
Looks good to me, if the tests pass!

omc3/plotting/utils/colors.py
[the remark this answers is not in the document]
fsoubelet: Got it, only `mpl` in there as of my last commit.

JoschD (approved):

PR #531, opened 2025-08-11
Tightening the screws, part 3

fsoubelet:
With this PR the last `Ruff` rules were activated and fixed, apart from the `NPY` one (which will be its own beast).

Also added the commit SHA from the previous linting PR (part 2) to the `.git-blame-ignore-revs` file.

Once again a lot of files changed but most changes are minimal. Worthy of note (and that one was a PITA): there are no more `os` shenanigans for path operations in the codebase :)

Also worthy of note: I've left commented out the previous version of the `os.path.whatever` we were using as I replaced with `pathlib`, I will delete them just before merging.


fsoubelet:
Also worth noting: the `Ruff` workflow we have runs checks on `omc3` but I have also checked and fixed the `tests` directory. Easily verified locally. Might want to check it in the workflow as well?

omc3/plotting/utils/annotations.py
```diff
@@ -26,7 +26,7 @@
     "alpha":              r'$\alpha_{{{0}}} \quad [m]$',
     "beta":               r'$\beta_{{{0}}} \quad [m]$',
     "betabeat":           r'$\Delta \beta_{{{0}}} \; / \; \beta_{{{0}}}$',
-    "betabeat_permile":   r'$\Delta \beta_{{{0}}} \; / \; \beta_{{{0}}} [$' '\u2030'r'$]$',
+    "betabeat_permil":   r'$\Delta \beta_{{{0}}} \; / \; \beta_{{{0}}} [$' '\u2030'r'$]$',  # noqa: ISC001 (wtf JoschD)
```
jgray-19, line 29: Not sure I agree with the noqa here... Why?

omc3/scripts/write_madx_macros.py
```diff
@@ -57,10 +57,9 @@ def define_observation_points_macros(list_of_bpms):
     return macro
 
 
-def tracking_macros(list_of_bpms, outdir):
-    obs_macro_file = outdir / OBS_POINTS
-    with open(obs_macro_file, 'w') as obs_script:
-        obs_script.write(define_observation_points_macros(list_of_bpms))
+def tracking_macros(list_of_bpms, outdir: Path | str):
```
jgray-19, line 60: Why no type on `list_of_bpms`

omc3/utils/logging_tools.py
```diff
@@ -355,36 +353,39 @@ def get_my_logger_name():
...
         caller_frame = caller_frame.f_back
         (caller_file_full, _, _, _, _) = inspect.getframeinfo(caller_frame)
-        caller_file, _ = os.path.splitext(caller_file_full)
-    return caller_file
+        caller_file = Path(caller_file_full).with_suffix("")
+    return str(caller_file)
 
 
-def _get_current_module(current_file=None):
+def _get_current_module(current_file=None) -> str:
```
jgray-19, line 369: No typing on current_file?

omc3/utils/logging_tools.py
```diff
@@ -355,36 +353,39 @@ def get_my_logger_name():
...
-    repo_parts = os.path.abspath(
-                    os.path.join(os.path.dirname(os.path.abspath(__file__)), os.path.pardir)
-                    ).split(os.path.sep)
 
+    # path_parts = os.path.abspath(current_file).split(os.path.sep)
+    # repo_parts = os.path.abspath(
+    #     os.path.join(os.path.dirname(os.path.abspath(__file__)), os.path.pardir)
+    # ).split(os.path.sep)
+    path_parts: tuple[str] = Path(current_file).resolve().parts
+    repo_parts = Path(__file__).resolve().parent.parent.parts
```
jgray-19, line 379: Are the comments here in case we need revert?

omc3/hole_in_one.py
```diff
@@ -408,19 +410,20 @@ def _replicate_harpy_options_per_file(options):
...
-    suffix = options.suffix or ""
+    dir_name: Path = Path(options.files).parent
+    file_name: str = Path(options.files).name
+    suffix: str = options.suffix or ""
 
     # Single bunch ---
     if tbt_data.nbunches == 1:
         if suffix:
-            options.files = join(dir_name, f"{file_name}{suffix}")
+            options.files = str(dir_name / f"{file_name}{suffix}")
```
jgray-19, line 426: Why do we always ensure it is a string?

omc3/madx_wrapper.py
```diff
@@ -93,9 +93,13 @@ def run_file(input_file: Path, output_file: Path = None, log_file: Path = None,
                madx_path=madx_path, cwd=cwd)
 
 
-def run_string(input_string: str,
-               output_file: Path = None, log_file: Path = None,
-               madx_path: Path = MADX_PATH, cwd: Path = None):
+def run_string(
+    input_string: str,
+    output_file: Path = None,
```
jgray-19, line 98: I thought you preferred `Path | None`??

tests/unit/test_plot_spectrum.py
```diff
@@ -77,25 +77,28 @@ def test_combined_bpms_stem_plot(tmp_path, file_path, bpms):
 
 
 @pytest.mark.basic
-def test_no_tunes_in_files_plot(tmp_path, file_path, bpms):
-    from glob import glob
+def test_no_tunes_in_files_plot(tmp_path: Path, file_path: Path, bpms: tuple[str, str]):
+
+    # for f in file_path.parent.glob("*"):
+        # print(f)
```
jgray-19, line 83: Is this some leftover debugging?

omc3/hole_in_one.py
fsoubelet: For safety. The previous implementation was always giving back strings, and the `opts` used to always pass strings around. For configuration writing purposes it's also nicer to have a string than some object.

I kept that behaviour unchanged.

omc3/utils/logging_tools.py
fsoubelet: See below.

omc3/model/accelerators/lhc.py
```diff
@@ -328,8 +330,10 @@ def important_phase_advances(self) -> list[list[str]] | None:
             return [["MKD.O5L6.B1", "TCTPH.4L1.B1"], ["MKD.O5L6.B1", "TCTPH.4L5.B1"]]
         return None
 
-    def get_synch_BPMs(self, index):
+    def get_synch_bpms(self, index: np.ndarray):
         # expect passing index.to_numpy()
+        # TODO: couldn't the following just be:
+        # np.isin(index, self.model.loc[f"BPMSW.33L2.B{self.beam}":].index)
```
fsoubelet, line 336: Eager to get your opinion @jgray-19 @JoschD

jgray-19: Actually the beam 2 has `33R8` not `33L2`

JoschD: a ) 
I am not sure if we need to use numpy here anyway, if we do not want to return an array. I'd say list-comprehension and `isin` are both readable.

b)
I'd be fine with:

select the BPM by beam and then do the 
```
return np.isin(index, self.model.loc[sync_bpm_start:].index).tolist()
```

c) 
More important is a doc-string in my opinion for this function.

d) 
This function is not used at all anyway.
We can also just remove it...

JoschD: Fairly sure it was replaced by https://github.com/pylhc/omc3/blob/4e59cd0f7eb6a4279dc3f5cfbaed2b03b1918df3/omc3/harpy/clean.py#L211

fsoubelet: Ah yes probably. In that code though the `np.isin` should be very justified. Anything that speeds up `harpy` is a win imo

JoschD: I doubt it would speed anything up, though. Resynching the bpms once at the beginning is not really a bottleneck

omc3/madx_wrapper.py
```diff
@@ -133,22 +137,25 @@ def _run(full_madx_script, log_file=None, output_file=None, madx_path=MADX_PATH,
 def _check_log_and_output_files(output_file, log_file):
     """Check read/write access. The EAFP way."""
     if output_file is not None:
-        open(output_file, "a").close()
+        with Path(output_file).open("a"):
+            pass
     if log_file is not None:
-        open(log_file, "a").close()
+        with Path(log_file).open("a"):
+            pass
```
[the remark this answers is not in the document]
fsoubelet: Not that I disagree, but this was the linter's recommended way iirc

JoschD (changes requested):

omc3/scripts/bad_bpms_summary.py
```diff
@@ -301,9 +301,9 @@ def read_harpy_bad_bpms_file(svd_file: Path) -> tfs.TfsDataFrame:
         tfs.TfsDataFrame: TfsDataFrame with all unique bad-bpms.
 
     """
-    TO_IGNORE = (HarpyReasons.NOT_IN_MODEL, )
-    TO_MARK = (HarpyReasons.KNOWN, )
-    COMMENT = "#"
+    to_ignore = (HarpyReasons.NOT_IN_MODEL, )
+    to_mark = (HarpyReasons.KNOWN, )
+    comment = "#"
```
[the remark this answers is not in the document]
fsoubelet: Ah, perfect. Linter complained about capitalized variable names in function body and I didn't think of `Final` to convey the constantness.

Done.

PR #544, opened 2025-08-13
Adding Linting Commit to Git-Rev File and Removing CodeClimate Badges

fsoubelet:
Adding the previous master commit's hash to `.git-blame-ignore-revs` and removing the code climate badges from `README.md`.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -4,3 +4,4 @@
 # Add below, one per line, SHAs of commits corresponding to "formatting" / "revisions"
 820a51e3c67c8675880b9c949ac09f4db8bd8e9f
 4e59cd0f7eb6a4279dc3f5cfbaed2b03b1918df3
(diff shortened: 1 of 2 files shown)


.git-blame-ignore-revs
```diff
@@ -3,4 +3,5 @@
 
 # Add below, one per line, SHAs of commits corresponding to "formatting" / "revisions"
 820a51e3c67c8675880b9c949ac09f4db8bd8e9f
-4e59cd0f7eb6a4279dc3f5cfbaed2b03b1918df3
\ No newline at end of file
+4e59cd0f7eb6a4279dc3f5cfbaed2b03b1918df3
+0c19f7f7aa2e69a46c0e72fb99d4c3a29a1726c1
```
JoschD, line 7: This does not look correct!

JoschD (approved):
LGTM

PR #553, opened 2025-10-17
Add the error on the calculation of the global correction.

jgray-19:
No one asked for this, but it was quite a simple implementation and is very informative to me and potentially others. 

Still a draft as the test on the error is just testing if the error exists, not actually testing that the error is calculated correctly. 

Also, ruffed these files.


jgray-19:
I will make a better pull request later that is far easier to read.

PR #580, opened 2026-05-30
Action calculation / rescaling fix

### LINKED ISSUE #530: [Bug]: Action calculation of LHCB2 is wrong
### Bug Description

2Jx and 2Jy columns in the Beam2 kick_x and kick_y return obviously wrong values for the action. Beam1 output is perfect.

The sqrt 2Jx and sqrt2Jy columns are correct, and do not match the values in 2jx and 2jy columns.

Seemed to work in Evan's simulation data, but not for the detuning studies done on 21st April. 2025.

fsoubelet:
This was triggered by #530, where some LHCB2 measurement highlighted inconsistency between the 2J and the sqrt(2J) columns, with the latter being correct.

I looked into the code in `omc3.optics_measurements.kick.py` and there were two things that stood out to me:

1. The `_get_action()` function computes sqrt(2J) and 2J as weighted averages with different data and weight functions. See:

| Quantity   | Weights (per BPM)                    |
|------------|--------------------------------------|
| `sqrt(2J)` | `beta / err_amps^2`                  |
| `2J`       | `beta^2 / (4 * amps^2 * err_amps^2)` |

Per BPM the values should always be consistent (for BPM `i` we fall back on `2J_i = sqrt(2J)_i^2` using the right side formulas), but the weighted means diverge when per-BPM actions vary.

2. The rescaling is inconsistent: `_rescale_actions()` multiplies both sqrt(2J) and 2J by the same `scaling_factor` variable. Considering the relationship and assuming the values for sqrt(2J) are correct (their rescaling too, then), we should be seeing `2J_RES = 2J * scaling_factor^2` but we are using `2J_RES = 2J * scaling_factor`.

Amplitude detuning analysis reads the rescaled `2J` column (aka 2J_RES) via `get_action_col()` which is how #530 stumbled onto this.

------

In this PR I propose to rebase our calculations on the observed correct `sqrt(2J)` values. This is done by:

1. Deriving `2J` from `sqrt(2J)` in `_get_action()` instead of its previous calculation using a different weighted mean.

2. Deriving the rescaled `2J` from the rescaled `sqrt(2J)` in `_rescale_actions()` instead of applying the wrong rescaling factor for this quantity. We could also just use the proper rescaling value but I've decided to go this way for consistency with the point 1 above.

-----

There was a new check added in the `test_optics_measurements` to ensure the consistency between the `2J` and `sqrt(2J)` in our output results. Same for the rescaled values and the corresponding errors.

The rest of the changes are type hints, an entry in the `CHANGELOG` and a patch version bump.

It would be good to run this branch's code on the analysis @emaclean mentioned in #530 so we can attest values for each column are alright.

Closes #530

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # OMC3 Changelog
 
+#### 2026-??-?? - v0.28.2 - _fsoubelet_
+
+- Fixed:
+    - Fixed an inconsistency in the rescaling for the action calculation leading to wrong values of 2J (while having correct values for sqrt(2J)), noticed in B2.
+
 #### 2026-03-27 - v0.28.1 - _fsoubelet_
 
 - Changed:
--- a/omc3/__init__.py
+++ b/omc3/__init__.py
@@ -11,7 +11,7 @@ omc3 is a tool package for the optics measurements and corrections group (OMC) a
 __title__ = "omc3"
 __description__ = "An accelerator physics tools package for the OMC team at CERN."
 __url__ = "https://github.com/pylhc/omc3"
-__version__ = "0.28.1"
+__version__ = "0.28.2"
 __author__ = "pylhc"
 __author_email__ = "pylhc@github.com"
 __license__ = "MIT"
--- a/omc3/optics_measurements/kick.py
+++ b/omc3/optics_measurements/kick.py
@@ -5,6 +5,7 @@ Kick
 This module contains kick functionality of ``optics_measurements``.
 It provides functions to compute kick actions.
 """
+
 from __future__ import annotations
 
 from contextlib import suppress
@@ -45,7 +46,9 @@ if TYPE_CHECKING:
     from omc3.optics_measurements.data_models import InputFiles
 
 
-def calculate(measure_input: DotDict, input_files: InputFiles, scale, header_dict: dict, plane):
+def calculate(
+    measure_input: DotDict, input_files: InputFiles, scale: float, header_dict: dict, plane: str
+) -> np.ndarray:
     """
 
     Args:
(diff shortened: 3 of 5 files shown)


fsoubelet:
As an aside, the tests for this are green but we have two failures from the `madng` response matrix correction. Could be due to a new version in `MAD-NG`? In any case, that's (very) unrelated code.

Maybe @jgray-19 you could look into that part.

fsoubelet:
Also forgot to mention in the original post: the rescaling is done with a factor determined as the weighted mean of the computed `"beta_hase" / "beta_amplitude"`.

It is possible that the calibration factors are not as good in B2. Then this ratio of beta phase to beta amp is further away from 1, and that feeds into the rescaling factor.

Because the faulty (previous) code was rescaling both quantities with the same ratio (aka computing `2J * r` instead of `2J * r^2`) the discrepancy is `|2J * r - 2J * r^2| = 2J * |r| * |1 - r|` so if this r gets away from 1 we notice the error.

So in short:
- B1 good calibration factors, r~1 and the error in the rescaling formula is negligible.
- B2 worse calibration factors, r!=1 and the error from the wrong rescaling becomes significant.

It's possible that this is not the reason but it's my best guess.

jgray-19:
I can do a proper review beforehand, but could you reanalyse the data Ewen was pointing to in the issue to demonstrate if this did fix the issue?

fsoubelet:
> I can do a proper review beforehand, but could you reanalyse the data Ewen was pointing to in the issue to demonstrate if this did fix the issue?

Sure, I will try to get the data from @emaclean. If you have a look at this code there should be no doubt we get the consistency between sqrt(2J) and 2J now. These changes do not touch anything about potentially getting wrong action altogether. See my "comment" above yours for my take on why this showed up for B2.

Related, I think it could be good to add another simple test analyzing simulation data and checking that the action value is correct. Will try to look into this.

jgray-19 (dismissed):
No real comments, looks good!

fsoubelet:
Had to include a fix so the test suite could run. @jgray-19 if you're still happy let's merge. I have a deprecation warning fix coming after this.

jgray-19 (approved):

PR #581, opened 2026-06-15
Deprecation and warning fixes

fsoubelet:
This branch fixes a few of the warnings and deprecations we could see showing up in the test runs.

- Fixes the deprecation of `scipy.odr` in favor of using `odrpack` as they recommend. Already had a previous PR on this, here's the end of it. Mostly adapting order of args, call sites etc.
- Fixes the "dataframe is highly fragmented" `PerformanceWarning` from pandas in the `harpy` code. This is due to modifying the linfiles a lot when building them (lots of column additions, rows additions etc). Fixed with a simple `frame.copy()` call at the right place (as suggested in the warning).
- "Fixes" the warning from `matplotlib` about the number of open figures. Since this does not happen in prod (we don't trigger all code paths in a go) I have just suppressed the warning in the `pytest` configuration. We could put some `plt.close("all")` in the tests but it feels rather ugly and destructive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # OMC3 Changelog
 
-#### 2026-??-?? - v0.28.2 - _fsoubelet_
+#### 2026-06-19 - v0.28.2 - _fsoubelet_
 
 - Fixed:
     - Fixed an inconsistency in the rescaling for the action calculation leading to wrong values of 2J (while having correct values for sqrt(2J)), noticed in B2.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 - Fixed:
     - Fixed an inconsistency in the rescaling for the action calculation leading to wrong values of 2J (while having correct values for sqrt(2J)), noticed in B2.
+    - Fixed a deprecation warning from `scipy.odr`.
+    - Fixed some `pandas` `PerformanceWarning`s in `omc3.harpy`.
+    - Fixed some type hints and docstrings in the `omc3.optics_measurements` modules.
 
 #### 2026-03-27 - v0.28.1 - _fsoubelet_
 
--- a/omc3/optics_measurements/beta_from_phase.py
+++ b/omc3/optics_measurements/beta_from_phase.py
@@ -143,7 +143,7 @@ def write(beta_df: pd.DataFrame, header: dict[str, Any], outputdir: str|Path, pl
 
 def n_bpm_method(
     meas_input: DotDict,
-    phase: pd.DataFrame,
+    phase: PhaseDict,
     plane: str,
     meas_and_mdl_tunes: tuple[float, float]
 ) -> tuple[tfs.TfsDataFrame, str]:
@@ -171,7 +171,7 @@ def n_bpm_method(
 
     Args:
         meas_input: `OpticsInput` object with optics CLI options.
-        phase: phase matrices of measurement with errors and model tfs (bpm x bpm).
+        phase: `PhaseDict` of phase matrices of measurement with errors and model tfs (bpm x bpm).
         plane: marking the horizontal or vertical plane, **X** or **Y**.
         meas_and_mdl_tunes: measured and model tunes.
 
@@ -761,7 +761,7 @@ def _get_header(header_dict: dict[str, Any], error_method: str, range_of_bpms: i
 
 def three_bpm_method(
     meas_input: DotDict,
-    phase: pd.DataFrame,
+    phase: PhaseDict,
     plane: str,
     meas_and_mdl_tunes: tuple[float, float],
 ) -> pd.DataFrame:
@@ -821,7 +821,7 @@ def three_bpm_method(
 
     Args:
         meas_input: `OpticsInput` object with optics CLI options.
-        phase: phase matrices of measurement with errors and model tfs (bpm x bpm).
+        phase: `PhaseDict` of phase matrices of measurement with errors and model tfs (bpm x bpm).
         plane: marking the horizontal or vertical plane, **X** or **Y**.
         meas_and_mdl_tunes: measured and model tunes.
 
--- a/omc3/optics_measurements/beta_from_phase.py
+++ b/omc3/optics_measurements/beta_from_phase.py
@@ -897,11 +897,14 @@ def three_bpm_method(
         sin_quadrup_model = tilted_errmeas**2 / np.power(np.sin(tilted_model), 4) * betmdl
 
     # Square to get variance terms (summed in quadrature across the two BPMs of each triplet)
+    # These terms can be zero when model phase advances are multiples of pi (e.g. in SPS) so
+    # we wrap it into the np.errstate context.
     sin_squared_model = np.square(sin_squared_model)
-    sin_squ_model_shift1 = sin_squared_model + np.roll(sin_squared_model, -1, axis=0) / np.square(cot_phase_model_shift1)
-    sin_squ_model_shift2 = sin_squared_model + np.roll(sin_squared_model, -2, axis=0) / np.square(cot_phase_model_shift2)
-    sin_quad_model_shift1 = sin_quadrup_model + np.roll(sin_quadrup_model, -1, axis=0) / np.square(cot_phase_model_shift1)
-    sin_quad_model_shift2 = sin_quadrup_model + np.roll(sin_quadrup_model, -2, axis=0) / np.square(cot_phase_model_shift2)
+    with np.errstate(divide="ignore", invalid="ignore"):
+        sin_squ_model_shift1 = sin_squared_model + np.roll(sin_squared_model, -1, axis=0) / np.square(cot_phase_model_shift1)
+        sin_squ_model_shift2 = sin_squared_model + np.roll(sin_squared_model, -2, axis=0) / np.square(cot_phase_model_shift2)
+        sin_quad_model_shift1 = sin_quadrup_model + np.roll(sin_quadrup_model, -1, axis=0) / np.square(cot_phase_model_shift1)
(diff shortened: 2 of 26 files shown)


fsoubelet:
I love scope creep!

The commits from today add some type hinting and docstring info in the modules of `omc3.optics_measurements`. Judge it as a whole, not for each to be 100% exact (we know from ty and pyrefly there were plenty of interface mismatches from the get go in this codebase). Let's say it has to make sense, although most if it should be fully correct.

Note that no actual code was changed, except for the last function in `omc3.optics_measurements.iforest` that I cleaned up a tiny bit.

omc3/optics_measurements/beta_from_phase.py
```diff
@@ -897,11 +897,14 @@ def three_bpm_method(
         sin_quadrup_model = tilted_errmeas**2 / np.power(np.sin(tilted_model), 4) * betmdl
 
     # Square to get variance terms (summed in quadrature across the two BPMs of each triplet)
+    # These terms can be zero when model phase advances are multiples of pi (e.g. in SPS) so
```
jgray-19, line 900: This doesn't feel right... Do we warn?

omc3/optics_measurements/coupling.py
```diff
@@ -64,20 +64,22 @@ def calculate_coupling(
...
+    tune_dict: dict[str, dict[str, float]],
     header_dict: dict,
 ) -> None:
     """
-    Calculates the coupling RDTs f1001 and f1010, as well as the closest tune approach Cminus (|C-|).
-    This represents the "2 BPM method" in https://cds.cern.ch/record/1264111/files/CERN-BE-Note-2010-016.pdf
-    (a more up-to-date reference will come in the near future).
+    Calculates the coupling RDTs f1001 and f1010, as well as the closest tune approach
+    Cminus (|C-|). This represents the "2 BPM method" in
+    https://cds.cern.ch/record/1264111/files/CERN-BE-Note-2010-016.pdf.
```
jgray-19, line 73: No up-to-date reference eh :D

omc3/optics_measurements/dispersion.py
```diff
@@ -210,7 +214,7 @@ def _calculate_normalised_dispersion_3d(meas_input: DotDict, input_files: InputF
     return output_df
 
 
-def _calculate_dp(model, disp, plane):
+def _calculate_dp(model: pd.DataFrame, disp: pd.DataFrame, plane: str) -> pd.Series:
     _m = "meas"
     df = pd.DataFrame(model).loc[:, ['S', f"MU{plane}", f"DP{plane}", f"D{plane}",
```
jgray-19, line 219: Did you not run ruff on this?

omc3/optics_measurements/dpp.py
```diff
@@ -36,8 +37,10 @@ def arrange_dpps(dpps: Sequence[float], tolerance: float = DPP_BIN_TOLERANCE):
     zero_offset = np.mean(_values_in_range(_find_range_with_element(ranges, closest_to_zero), dpps))
     LOGGER.debug(f"dp/p closest to zero is {dpps[closest_to_zero]}")
     if np.abs(zero_offset) > tolerance:
-        LOGGER.warning(f"Analysed files have large momentum deviation {zero_offset}. "
-                       f"Optics parameters might be wrong.")
+        LOGGER.warning(
+            f"Analysed files have large momentum deviation {zero_offset}. "
+            "Optics parameters might be wrong."
```
jgray-19, line 58: I got this warning recently, why does it say optics parameters might be wrong?

omc3/optics_measurements/dpp.py
```diff
@@ -75,30 +80,53 @@ def _compute_ranges(dpps: Sequence[float], tolerance: float) -> list[list[int]]:
...
 
-def append_amp_dpp(list_of_tfs: Sequence[tfs.TfsFile], dpp_values: Sequence[float]):
-    """ Add the dpp values to the DPP-header of the tfs files, if larger than the DPP-tolerance, otherwise set to zero.
-    This is intended to the DPP value for on-momentum files to zero. """
+def append_amp_dpp(
+    list_of_tfs: Sequence[tfs.TfsDataFrame], dpp_values: Sequence[float]
+) -> Sequence[tfs.TfsDataFrame]:
+    """
+    Add the dpp values to the DPP-header of the tfs files, if larger than the DPP-tolerance,
+    otherwise set to zero. This is intended to the DPP value for on-momentum files to zero.
```
jgray-19:
```suggestion
    otherwise set to zero. This is intended to set the DPP value for on-momentum files to zero.
```

Did you mean this?

omc3/optics_measurements/iforest.py
```diff
@@ -77,7 +157,19 @@ def get_significant_features(bpm_tfs_data, data_for_clustering, bad_bpms, good_b
...
 
-def detect_anomalies(contamination, data, plane):
+def detect_anomalies(
+    contamination: float, data: pd.DataFrame, plane: str,
+) -> tuple[pd.DataFrame, pd.DataFrame, np.ndarray]:
+    """Fits an isolation forest and separates data into anomalous and normal BPMs.
+
+    Args:
+        contamination: expected proportion of outliers in the data.
+        data: normalized BPM feature data.
```
jgray-19, line 167: If you let me loose on this documentation at some point, I might very well convert all American spellings to British...

omc3/optics_measurements/interaction_point.py
jgray-19, line 44: Why None (Not mentioned here)?

omc3/optics_measurements/rdt.py
```diff
@@ -200,7 +200,28 @@ def _rdt_to_order_and_type(rdt: RDTTuple):
     return f"{rdt_type}_{orders[j + k + l + m]}"
 
 
-def _best_90_degree_phases(meas_input, bpm_names, phases, tunes, plane):
+def _best_90_degree_phases(
```
jgray-19, line 203: I feel like we do this in more than one place? Could add an issue...?

Also probably worth going through the issues to check for any stale?

CHANGELOG.md
jgray-19: If you make the changes today, I promise to review tomorrow morning!

jgray-19 (changes requested):
Very minor comments!

omc3/optics_measurements/beta_from_phase.py
fsoubelet: Well if I remove the `np.errstate` context I have added, a warning will display. However it is simply a "An invalid value was encountered in divide" which is not so informative.

Considering these terms feed into calculations for `beterr` (ERR BETA column) and `sin_quad_model_shift*` (ERR ALPHA column), in lines 910-915 below. Error estimates.

We have an EPSILON guard (lines 864-866) to protect he main `bet_frac` computation for beta function calculation.

I would say when a BPM triplet is degenerate (this is rare, for instance we never see this in the LHC case), the error estimate going to `inf` or `nan` is correct.

Short answer is I think this is okay behavior all around and I would prefer not to overwhelm the user with either a cryptic message or a very long explanation.

omc3/optics_measurements/dispersion.py
fsoubelet: You mean for formatting? Running a full formatting pass lead to gigantic changes in all files, some not really for the better as well (we do have some gigantic expressions here and there, which are legitimate).

Felt like the git diffs would give you an aneurism if I included it :D

omc3/optics_measurements/dpp.py
fsoubelet: Yes, thanks

omc3/optics_measurements/dpp.py
fsoubelet: This function re-centers the dp/p values so the bin closest to 0 becomes 0, but the `zero_offset` is the actual mean dp/p of that "on-momentum" bin.

You get this warning when `zero_offset` is large, which mean that even the "on-momentum" files were taken with significant momentum deviation from the design closed orbit. Or the optics parameters could be wrong.

I expect you would see this for very off-momentum measurements when we trim the RF. It's not the most informative message to log but the answer is a little long. Might include my answer in a comment there?

omc3/optics_measurements/iforest.py
fsoubelet: do_it_palpatine.jpg

omc3/optics_measurements/interaction_point.py
fsoubelet: Triggered if the accelerator class has no `get_ip` method, which is a few lines above.

omc3/optics_measurements/rdt.py
fsoubelet: not always exactly for 90 degrees. Yes a check after shutdown might be good.

CHANGELOG.md
fsoubelet: Bet!

omc3/optics_measurements/dpp.py
jgray-19: Could this re-centering of 0 affect the phase measurements, for example Mattia's MD?

fsoubelet: No, I don't think so. This is used in two ways:

- For filtering / grouping (`dpp_frames`, `dpp_frames_indices`) where the dp/p value is used to select which files belong to the same momentum bin.
- For numerical fitting in `dispersion.py` (triggers when we analyse on-off momentum measurements), but the shift only changes the intercept and not the fitted slope, which is the actual result of the dispersion.

jgray-19: Ok. So no optics function calculations are momentum dependent then?

jgray-19 (approved):

omc3/optics_measurements/dpp.py
fsoubelet: No, only the dispersion is.

PR #584, opened 2026-07-02
Use a different variable for getting energy, possibly will break GUI!!!!

jgray-19:
Bjorn told me not to use the HX:ENG variable. I think this might be the same source. 

I also saw in timber that every second the energy varied by about 0.12, so introduced an averaging technique to potentially get a more stable energy value (does not change every second). 

Potentially requires input from OP to understand if this is correct.


fsoubelet:
Are you ok putting this PR as `draft` until this is discussed with OP?

PR #585, opened 2026-07-03
[Harpy] Parallelisation Strategy

fsoubelet:
This PR introduces a parallelisation strategy for the `harpy` harmonic analysis.
Let me preface this by saying: no code that does any of the calculations has been touched.

This is a long PR text, I'll try to split it.

## Context

Currently, when performing harmonic analysis for the bunches in a file, `harpy` runs the analysis serially, bunch after bunch. We have a wall time of ~25-30s for analysing a bunch so this scales roughly as `n_bunches * 30 seconds`.

The operations within `harpy` (SVD, some big dot products, FFTs) dispatch hungrily across all available threads via `BLAS`.

As it turns out for our operations this is not a great idea, and benchmarks show that limiting `BLAS` to a single thread gives similar results (slightly faster for SVD, slightly slower for dot product, all within 5-10% in any case). This is an opportunity to free up the threads and parallelise differently.

## What is changed

The parallelisation strategy is changed. Instead of running bunch after bunch and letting `BLAS` eat up all threads unproductively, the following is done:
- The `harpy` workload of single bunches sets a hard cap for `BLAS` to one (1) thread,
- The bunches are now parallelised across workers.

## Implementation

1. Most of the new code is in a new `omc3.harpy._parallel` module, with heuristics to determine the number of workers to dispatch to. This is done based on estimated peak RSS for a given bunch (from the TbT data and from the `harpy` options), the detected number of available logical cores and RAM, and user input (more on that later).

2. The main function in `harpy` now calls this heuristic first then dispatches work. It performs the same thing as the current behaviour if `n_jobs=1`, and starts a `ProcessPoolExecutor` to parallelise across bunches if `n_jobs>1`.

3. A CLI flag `--n_jobs` is provided to give the user (and GUI) control over this strategy. Providing `0` (the default) leads to automatic determination as described above. Giving `1` leads to serial processing, which is the current behaviour. Providing `N>0` uses uses up to N workers, still bounded by the automatic safety limits (available cores, number of bunches, and RAM budget): an explicit `N` can only throttle the pool below the automatic choice, but never oversubscribe it.

Some little things:

- Version and CHANGELOG are bumped in preparation for a feature release. Would be nice to have #579 in as well, and maybe some further improvements in the calculation parts of `harpy` I am reserving for another PR.
- `threadpoolctl` is added to the dependencies (was already a transitive dependency),
- `psutil` is added to the dependencies.

### Performance Gains

The theoretical gains, assuming we are not limited by the resources of the machine, is a factor of `n_bunches` at most. In practice there is some overhead due to the dispatch of jobs but it is small, especially compared to the workload's runtime.

Previously: a file with `n_bunches` would analyse them one by one.
Now: a file with `n_bunches` will start all analyses simultaneously and be done in ~ the same time it takes to complete a single bunch's analysis.

## Tests

I have added unit tests for both the parallelisation heuristics, and for comparing serial results vs parallel workers results on the same inputs. The latter is done for linfiles but also the `.amps[xy]` and `.freqs[xy]` files from the `full_spectra` output (as full_spectra is our default in the GUI).

## Field Tests

I have tested this in the CCC starting workloads from the GUI. Reproducing previous analyses leads to the exact same results, down to the optics analysis downstream.

See below for GUI + harpy parallelisation considerations when we are dealing with multiple files.

### Wider parallelisation with GUI

There are two layers to the interplay here. This is assuming the default GUI setting of `Run Per-File Tasks in Parallel`.

When selecting multiple files in the BPM panel and starting the `Analyse spectra` command, the GUI issues a `python -m omc3.hole_in_one --harpy [...]` command for each file, instantly. This is done via `ssh` submission to the relevant optics server. What happens next is:

- _Currently_: All started processes (from GUI, one per file) start ~at once, and for each file the bunches are done sequentially. With a wall time of ~30s for analysing a bunch each file takes at minimum 30s x `n_bunches`. Considering `BLAS` maxes out the threads on the server, with a lot of files (or the newer 5 bunches / beam measurements) this can lead to throttling and additional slowdown on top of the `30s x n_bunches`.
- _This PR_: All started processes (from GUI, one per file) start ~at once, and each analyses the resources then parallelises across bunches within detected limits (available cores, RAM etc). If the resources are high enough then *all* bunches run in parallel and we are reduced to the min wall time of `~30s` for a given bunch's analysis.

The general outcomes:

- For a single file the gain is nice, since we always fall down to base wall time with our optics server resources: from `n_bunches x ~30s` -> `~30s`.
- For many files the gain can be even more, since the old behaviour can throttle the server and in that case we might see a better speed-up.
- *Importantly*, when analysing the kick files as they come in we can assume on would always analyse just one file at a time and be done in `~30s` every time. This lets analysis stay on top of the measurements considering we can only get a kick every minute (currently, especially with the 5 bunches the analysis falls behind measurements quite quickly).

### Benchmarks from the CCC

The python edge environment is currently on this branch. I have reproduced some analyses from earlier in the year and compared the prod version's performance (current `omc3`) to the edge version (this branch).

All results down to the optics analysis are perfectly identical. Below are some screenshots of the performance gains, for the 5 bunches measurements.

|                     | Current omc3 | This Branch omc3 |
| ------------------- | ------------ | ---------------- |
| Optics Server Usage |prod_usage | edge_usage |
| Performance Logs    | prod_logs | edge_logs |
| Total Wall Time     | 120 seconds | 31 seconds |

**Verdict**: we now finish all much faster and close to the min wall time for one bunch analysis, and we utilise much less of the machine resources, which means we have room for a lot more parallelisation (more files, more bunches).

Notice also that the new `harpy` is a little more talkative in the logs :)

Results of the analyses are the exact same, see the final optics for instance:

results_comparison

For more benchmark details see the OMC logbook entries of today, `2026-07-03`.

### DIFF
--- a/omc3/harpy/_parallel.py
+++ b/omc3/harpy/_parallel.py
@@ -197,7 +197,7 @@ def decide_n_workers(
 
     # Compute the number of workers / jobs to dispatch, log it and return
     n_jobs: int = max(1, min(cores_cap, n_bunches, ram_cap))
-    LOGGER.info(
+    LOGGER.debug(
         f"Parallel harpy: n_jobs={n_jobs} (cores_cap={cores_cap}, bunches={n_bunches}, "
         f"ram_cap={ram_cap_str}, peak/bunch~{peak_rss / 1e9:.1f}GB, available~{available_ram_str})"
     )
--- a/omc3/harpy/_parallel.py
+++ b/omc3/harpy/_parallel.py
@@ -42,8 +42,9 @@ SAFETY_MARGIN = 0.85  # 85% to leave room for other processes etc.
 # this value is safe (leads to fewer workers for parallelisation and no risk of bein OOM).
 _BASELINE_RSS_BYTES = int(1.5e9)  # 1.5GB peak RSS for a bunch in the lowest consuming scenario
 
-# Generous upper bound on the number of frequency lines selected by the narrow
-# (resonance-band) mask; its transient is negligible next to the baseline.
+# Generous upper bound on the number of frequency lines selected by the narrow mask.
+# Its transient is negligible next to the baseline. See comments inside of the
+# estimate_peak_rss_bytes_per_bunch function for details and how this plays a role.
 _NARROW_MASK_BINS = 1 << 14  # equivalent to 2**14
 
 # ----- System information ----- #
--- a/omc3/harpy/_parallel.py
+++ b/omc3/harpy/_parallel.py
@@ -37,7 +37,7 @@ _BYTES_COMPLEX128 = 16
 
 # Fraction of available RAM the worker pool is allowed
 # to use, with a little bit of margin
-SAFETY_MARGIN = 0.9
+SAFETY_MARGIN = 0.85
 
 # Empirically measured floor on the optics1 server (64-core, 536 BPMs, lin-only path):
 # Python + NumPy + libraries + one bunch of data + the SVD working set. Overestimating
@@ -140,3 +140,66 @@ def estimate_peak_rss_bytes_per_bunch(harpy_input: DotDict, n_bpms: int) -> int:
 
     # Return the estimated peak + minimum memory occupancy (data loading etc)
     return _BASELINE_RSS_BYTES + transient
+
+
+def decide_n_jobs(
+    harpy_input: DotDict,
+    n_bunches: int,
+    n_bpms: int,
+    *,
+    requested: int = 0,
+    margin: float = SAFETY_MARGIN,
+) -> int:
+    """
+    Choose the number of worker processes for per-bunch analysis. Each bunch is sent
+    to a worker. This assumes the number of BLAS threads is limited (for `numpy.svd`
+    operations for instance where it maxes out the threads and doesn't do much with
+    it).
+
+    Note
+    ----
+        The `requested` argument mirrors the `n_jobs` for the harpy parameters. Passing
+        `0` selects automatically (all usable cores, RAM permitting); `1` runs serially
+        (which is the old behaviour) and providing an integer value `N` caps the pool at
+        either `N` processes or the automatically determined number (RAM permitting).
+
+        The result is `max(1, min(cores_cap, n_bunches, ram_cap))`, where `ram_cap` is
+        `floor(margin * available_ram / peak_rss_per_bunch)` (or unconstrained when the
+        available RAM could not be determined).
+
+    Args:
+        harpy_input (DotDict): Harpy analysis settings, forwarded to
+            :func:`estimate_peak_rss_bytes_per_bunch` for the per-bunch RSS estimate.
+        n_bunches (int): Number of bunches to process. Caps the pool since each bunch
+            goes to at most one worker.
+        n_bpms (int): Number of BPMs (use the pre-clean count as a safe upper bound),
+            used for the per-bunch RSS estimate.
+        requested (int): the `n_jobs` option passed to ``harpy``. See the note above.
+        margin (float): fraction of available RAM the worker pool may occupy, leaving
+            the remainder as headroom against under-estimating the peak and for other
+            running processes.
+
+    Returns:
+        The number of worker processes to use, always at least 1.
+    """
+    # Determine a few parameters
+    cores_cap: int = requested if requested > 0 else usable_cores()
+    peak_rss: int = estimate_peak_rss_bytes_per_bunch(harpy_input, n_bpms)
+    available_ram = available_ram_bytes()
+
+    # Determine the available RAM to use, including the safety margin
+    # If unknown RAM -> do not let it constrain the choice
+    if available_ram is None:
+        ram_cap: int = n_bunches
+        available_ram_str, ram_cap_str = "unknown", "n/a"
+    else:
+        ram_cap = int(margin * available_ram / peak_rss)
+        available_ram_str, ram_cap_str = f"{available_ram / 1e9:.0f}GB", str(ram_cap)
+
+    # Compute the number of workers / jobs to dispatch, log it and return
+    n_jobs: int = max(1, min(cores_cap, n_bunches, ram_cap))
+    LOGGER.info(
+        f"Parallel harpy: n_jobs={n_jobs} (cores_cap={cores_cap}, bunches={n_bunches}, "
+        f"ram_cap={ram_cap_str}, peak/bunch~{peak_rss / 1e9:.1f}GB, available~{available_ram_str})"
+    )
+    return n_jobs
--- a/omc3/harpy/_parallel.py
+++ b/omc3/harpy/_parallel.py
@@ -91,7 +91,7 @@ def available_ram_bytes() -> int | None:
     """
     Returns currently available RAM in bytes (memory that can be
     given to processes without swapping), via ``psutil``. Returns
-    or ``None`` if it cannot be determined.
+    ``None`` if it cannot be determined.
     """
     try:
         return psutil.virtual_memory().available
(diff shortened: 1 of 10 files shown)


omc3/harpy/_parallel.py
```diff
@@ -0,0 +1,227 @@
+"""
+Parallel
```
jgray-19: I think maybe this title could be more descriptive

omc3/harpy/_parallel.py
```diff
@@ -0,0 +1,227 @@
...
+LOGGER: Logger = logging_tools.get_logger(__name__)
+
+# ----- Some constants ----- #
+
+_BYTES_FLOAT64 = 8
+_BYTES_COMPLEX128 = 16
+
+# Fraction of available RAM the worker pool is allowed to use,
+# with a little bit of margin. Could be tweaked.
+SAFETY_MARGIN = 0.85  # 85% to leave room for other processes etc.
```
jgray-19: Is this you only use 15% of the machine or 85%?

omc3/harpy/_parallel.py
```diff
@@ -0,0 +1,227 @@
...
+
+    Returns:
+        The number of worker processes to use, always at least 1.
+    """
+    # Determine a few parameters. A requested N only ever caps *below* the usable cores:
+    # workers are CPU-bound (BLAS capped to 1 thread), so more workers than cores just
+    # oversubscribes with no gain. Hence N can be throttled down but never oversubscribe.
+    cores_cap: int = min(requested, usable_cores()) if requested > 0 else usable_cores()
+    peak_rss: int = estimate_peak_rss_bytes_per_bunch(harpy_input, n_bpms)
+    available_ram: int | None = available_ram_bytes()
```
jgray-19, line 204: Why is typing needed here?

omc3/harpy/_parallel.py
```diff
@@ -0,0 +1,227 @@
...
+
+    # Determine the available RAM to use, including the safety margin
+    # If unknown RAM -> do not let it constrain the choice
+    if available_ram is None:
```
jgray-19, line 211: If available ram is none, isn't it best to not be parallel at all?

pyproject.toml
```diff
@@ -58,6 +58,8 @@ dependencies = [
   "tables >= 3.10.1",
   "requests >= 2.27",
   "odrpack >= 0.4.0",
+  "psutil >= 7.0",
```
jgray-19, line 61: Maybe better to have this as optional?

tests/unit/test_harpy_parallel.py
```diff
@@ -0,0 +1,197 @@
...
+        raise NotImplementedError
+
+    monkeypatch.delattr(os, "process_cpu_count", raising=False)  # simulate lower Python
+    monkeypatch.delattr(os, "sched_getaffinity", raising=False)  # simulate non-Linux
+    monkeypatch.setattr(_parallel.psutil, "Process", lambda: SimpleNamespace(cpu_affinity=_raise))
+    monkeypatch.setattr(os, "cpu_count", lambda: 7)  # patch as well as we can't predict the CI machine
+    assert _parallel.usable_cores() == 7
+
+
+# ----- Estimation of Peak RSS per Bunch ----- #
```
jgray-19, line 82: These tests seem excessive

tests/unit/test_harpy_parallel.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+# ----- Determination of Number of Parallel Workers ----- #
+
+
+@pytest.mark.basic
+def test_decide_n_jobs_core_bound(monkeypatch):
+    """So much RAM the number of cores limits us."""
+    monkeypatch.setattr(_parallel, "usable_cores", lambda: 8)  # 8 cores only
+    monkeypatch.setattr(_parallel, "available_ram_bytes", lambda: int(1e12))  # 1000GB of RAM!
```
jgray-19, line 134: Where did you download this from?!

jgray-19 (commented):
Looks good, nice PR!

omc3/harpy/_parallel.py
fsoubelet: Do you mean the header? It won't be part of the documentation as the module is private but I could make it like "Parallel Heuristics" or something?

omc3/harpy/_parallel.py
fsoubelet: see below, it means we allow up to 85%. This is what I consider the `ram_cap` later on. I try to leave some margin as since the GUI would start several files in quick succession, it is possible that we "overestimate" how much ram we can use if the various file processes all try to do this estimate at the same time. Very unlikely that it's the exact timing but it can be 2 files for instance. I like having a margin.

That being said, if it's just RAM the processes would throttle a bit but no biggie.

omc3/harpy/_parallel.py
fsoubelet: Typing is never *needed* but I like to have it :)

omc3/harpy/_parallel.py
fsoubelet: Well the most likely scenario where `available_ram` is `None` is that querying the info didn't work, which does not necessarily mean it will constrain us. But I see your point. I thought it was simpler this way but we could also defer to 1 job.

pyproject.toml
fsoubelet: Hmm since I feel like `harpy` is default functionality of `omc3` and the automatic parallelism is default behaviour, it would be strange that a dependency required for this behaviour is missing when someone goes `python -m pip install omc3`.

tests/unit/test_harpy_parallel.py
fsoubelet: I wanted code coverage :/

tests/unit/test_harpy_parallel.py
fsoubelet: Damn it took me a while to understand this comment lmao

I will tell you but only in private.

pyproject.toml
jgray-19: I accept if these are small packages

omc3/harpy/_parallel.py
jgray-19: I think the lowest risk method is that if we don't know how much memory someone has, we assume they have minimal.

omc3/harpy/_parallel.py
jgray-19: Could you make the comment clearer or the variable name potentially?

omc3/harpy/_parallel.py
jgray-19: `HarPy Parallel Helper`?

fsoubelet: Sure!

omc3/harpy/_parallel.py
fsoubelet: Will do!

pyproject.toml
fsoubelet: Well the complete `tar` for `psutil` is about 500kB which is quite small. It's also a very popular and relied on package from the Python ecosystem so I would not be worried about it.

omc3/harpy/_parallel.py
fsoubelet: Okay I actually agree. However I think I would only do this in automatic determination mode:

- if `n_jobs = 0` (auto) -> we set `ram_cap = 1` and refuse to guess.
- if `n_jobs = N` (requested) -> we still honour `N` because I would assume the user knows their machine, which should be better than the automatic detection we attempt.

I think I could log a warning in both cases to ease diagnostics should someone wonder why they are not seeing the speed they might expect.

Would you agree?

jgray-19: I agree! The auto warning can indicate them to manually set the number of jobs, while the other says continuing with this many jobs even though I don't know your RAM

fsoubelet: Done

jgray-19 (approved):
Great!

PR #588, opened 2026-08-10
More Ruff 

fsoubelet:
In version `0.16.0` released a few weeks ago, Ruff has started enabling a much larger set of rules by default (413, up from 59): see [the release notes](https://github.com/astral-sh/ruff/releases/tag/0.16.0).

This PR adds some code changes to satisfy the new rules, so other PRs aren't blocked by the `checks` workflow.

**Note:** To satisfy the rules about timezone naive datetimes, I have enabled in a few places a default UTC timezone. This was only in places where we create a time stamp and add it to a file (from corrections for instance) and should not be an issue. Please check.

TODO after merging: adding the merge commit to `.git-blame-ignore-revs`.

### DIFF
--- a/omc3/amplitude_detuning_analysis.py
+++ b/omc3/amplitude_detuning_analysis.py
@@ -288,10 +288,12 @@ def analyse_with_bbq_corrections(opt: DotDict) -> tuple[TfsDataFrame, TfsDataFra
     return kick_df, bbq_df
 
 
-def get_kick_and_bbq_df(kick: Path | str, bbq_in: Path | str,
-                        beam: int = None,
-                        filter_opt: FilterOpts = None,
-                        ) -> tuple[TfsDataFrame, TfsDataFrame]:
+def get_kick_and_bbq_df(
+    kick: Path | str,
+    bbq_in: Path | str,
+    beam: int | None = None,
+    filter_opt: FilterOpts | None = None,
+) -> tuple[TfsDataFrame, TfsDataFrame]:
     """Load the input data."""
     bbq_df = None
     if bbq_in is not None and bbq_in == INPUT_PREVIOUS:
--- a/omc3/amplitude_detuning_analysis.py
+++ b/omc3/amplitude_detuning_analysis.py
@@ -508,10 +508,10 @@ def _get_bbq_data(beam: int, input_: Path | str | int, kick_df: TfsDataFrame) ->
                 t_start - t_delta, t_end + t_delta, keys=timber_keys, names=dict(zip(timber_keys, bbq_cols))
             )
         else:  # input_ is a file name or path
-            LOG.debug(f"Getting bbq data from file '{str(input_):s}'")
+            LOG.debug(f"Getting bbq data from file '{input_!s:s}'")
             data = read_timed_dataframe(input_)
             if not len(data.index):
-                raise ValueError(f"No entries in {str(input_):s}.") from e
+                raise ValueError(f"No entries in {input_!s:s}.") from e
 
     else:  # input_ is a number, assumed to be a fill number
         LOG.debug(f"Getting timber data from fill '{input_:d}'")
--- a/omc3/model/model_creators/abstract_model_creator.py
+++ b/omc3/model/model_creators/abstract_model_creator.py
@@ -499,9 +499,9 @@ class SegmentCreator(ModelCreator, ABC):
             madx_script += "\n".join(
                 [
                     f'call, file="{corrections_path}";',
-                    (f"exec, twiss_segment(forward_SbSSEQ, "
+                    ("exec, twiss_segment(forward_SbSSEQ, "
                     f'"{twiss_forward_corr_path}", biniSbSParams);'),
-                    (f"exec, twiss_segment(backward_SbSSEQ, "
+                    ("exec, twiss_segment(backward_SbSSEQ, "
                     f'"{twiss_backward_corr_path}", bendSbSParams);'),
                     "",
                 ]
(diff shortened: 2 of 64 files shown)


omc3/amplitude_detuning_analysis.py
```diff
@@ -508,10 +510,10 @@ def _get_bbq_data(beam: int, input_: Path | str | int, kick_df: TfsDataFrame) ->
                 t_start - t_delta, t_end + t_delta, keys=timber_keys, names=dict(zip(timber_keys, bbq_cols))
             )
         else:  # input_ is a file name or path
-            LOG.debug(f"Getting bbq data from file '{str(input_):s}'")
+            LOG.debug(f"Getting bbq data from file '{input_!s:s}'")
             data = read_timed_dataframe(input_)
             if not len(data.index):
-                raise ValueError(f"No entries in {str(input_):s}.") from e
+                raise ValueError(f"No entries in {input_!s:s}.") from e
```
jgray-19, line 516: What is this !s doing more than the automatic conversion from the f string?

omc3/model/model_creators/abstract_model_creator.py
```diff
@@ -502,10 +499,10 @@ def get_madx_script(self) -> str:
...
                 [
                     f'call, file="{corrections_path}";',
-                    f"exec, twiss_segment(forward_SbSSEQ, "
-                    f'"{twiss_forward_corr_path}", biniSbSParams);',
-                    f"exec, twiss_segment(backward_SbSSEQ, "
-                    f'"{twiss_backward_corr_path}", bendSbSParams);',
+                    (f"exec, twiss_segment(forward_SbSSEQ, "
+                    f'"{twiss_forward_corr_path}", biniSbSParams);'),
+                    (f"exec, twiss_segment(backward_SbSSEQ, "
+                    f'"{twiss_backward_corr_path}", bendSbSParams);'),
```
jgray-19, line 505: Aren't there some useless fstrings here?

omc3/plotting/spectrum/utils.py
```diff
@@ -318,7 +318,7 @@ def get_data_for_bpm(data: dict, bpm: str, rescale: bool) -> dict:
                 data_series[plane][AMPS] = rescale_amp(data_series[plane][AMPS])
 
             if any(data_series[plane][AMPS].isna()):
-                raise Exception("NAN FOUND")
+                raise Exception("NAN FOUND")  # ruff: ignore[TRY002]
```
jgray-19: Which rule is this?

omc3/plotting/utils/lines.py
```diff
@@ -25,7 +30,7 @@ class MarkerList:
     """Create a list of predefined markers."""
 
     # markers = ["s", "o", ">", "D", "v", "*", "h", "^", "p", "X", "<", "P"]  # matplotlib 2.++
-    markers = ["s", "o", ">", "D", "v", "*", "h", "^", "p", "<"]
+    markers: ClassVar[list[str]] = ["s", "o", ">", "D", "v", "*", "h", "^", "p", "<"]
```
jgray-19, line 33: What is this `ClassVar`?

omc3/segment_by_segment/propagables/alpha.py
```diff
@@ -46,8 +46,8 @@ def get_at(cls, names: IndexType, meas: OpticsMeasurement, plane: str) -> ValueE
     def in_measurement(cls, meas: OpticsMeasurement) -> bool:
         """ Check if the alpha phase is in the measurement data. """
         try:
-            meas.beta_phase_x
-            meas.beta_phase_y
+            _x = meas.beta_phase_x  # just testing presence via access | why not test alfx?
```
jgray-19, line 49: Another PR?

jgray-19 (dismissed):
Looks good to me, some of natures defaults are quite nice. My only concern is the use of `!s` in fstrings which I don't understand...?

omc3/amplitude_detuning_analysis.py
fsoubelet: This is from the `RUF010` rule, see https://docs.astral.sh/ruff/rules/explicit-f-string-type-conversion/

I honestly applied it to satisfy Ruff but we could add an ignore to the line, or just add the rule to ignores.

omc3/plotting/spectrum/utils.py
fsoubelet: TRY002, to dissuade raising a bare exception in this case

https://docs.astral.sh/ruff/rules/raise-vanilla-class/

omc3/plotting/utils/lines.py
fsoubelet: See the related commit (), and the rule https://docs.astral.sh/ruff/rules/mutable-class-default/

We have mutable class defaults which is not the best thing, but it would be a lot of work to rewrite so we don't. Annotating as `ClassVar` helps better classify and suppress the warning.

omc3/segment_by_segment/propagables/alpha.py
fsoubelet: I imagine it could be. I think at the time there was a thinking like "if beta_phase_x is present then the alpha function will be too" but memory is foggy

omc3/plotting/spectrum/utils.py
jgray-19: Why not use a ValueError?

omc3/amplitude_detuning_analysis.py
jgray-19: So 

`f'{astr}'` would fail? And you need `f'{astr!s}'`?

omc3/plotting/spectrum/utils.py
fsoubelet: Yes will do

jgray-19 (approved):
lgtm

omc3/amplitude_detuning_analysis.py
fsoubelet: The first one should be enough in a lot of cases but not all I believe. The second one is the equivalent to `str(something)` but more specific and recommended by the rule.

